PR #7, opened 2023-05-01
Inference

mvpatel2000:
Adds inference example for SD2 using our model. I think it's fine to not scrub paths given we're using this as a working directory + none of it is sensitive


Landanjs (approved):
shore

PR #12, opened 2023-05-06
LAION dataloader does not accept single string remote and local

Landanjs:
LAION dataloader does not accept single string remote and local arguments since these are sequences, but have difference lengths


mvpatel2000 (approved):
Should be valueerror not assert

PR #13, opened 2023-05-09
Pass attention mask to CLIP model

Landanjs:
Adds the attention mask as a return value in the LAION and COCO dataloaders. The attention mask is passed to the CLIP model in both `forward()` and `_prepare_text_embeddings()` functions. This will prevent the CLIP model from attending to the padded embeddings.

Convergence impact TBD

I did not add this to the LogDiffusionImage callback yet to keep the PR small. 

As of now, the attention mask passed to the U-Net forward is not used, but this PR seems to be close to adding it to the cross attention layers: https://github.com/huggingface/diffusers/pull/2634. I might try it 😁

### DIFF
--- a/diffusion/datasets/laion/laion.py
+++ b/diffusion/datasets/laion/laion.py
@@ -97,9 +97,10 @@ class StreamingLAIONDataset(StreamingDataset):
             padding='max_length',
             max_length=self.tokenizer.model_max_length,
             truncation=True,
-        )['input_ids']
-        tokenized_caption = torch.tensor(tokenized_caption)
-        out = {'image': img, 'captions': tokenized_caption}
+        )
+        input_ids = torch.tensor(tokenized_caption['input_ids'])
+        attention_mask = torch.tensor(tokenized_caption['attention_mask'])
+        out = {'image': img, 'captions': input_ids, 'attention_mask': attention_mask}
         if 'caption_latents' in sample:
             out['caption_latents'] = torch.from_numpy(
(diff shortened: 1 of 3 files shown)


diffusion/models/stable_diffusion.py
```diff
@@ -158,18 +158,23 @@ def forward(self, batch):
             latents, conditioning = batch[self.image_latents_key], batch[self.text_latents_key]
         else:
             inputs, conditioning = batch[self.image_key], batch[self.text_key]
+            # Do we need this?
             conditioning = conditioning.view(-1, conditioning.shape[-1])
+            attention_mask = None
+            if 'attention_mask' in batch:
+                attention_mask = batch['attention_mask'].view(-1, batch['attention_mask'].shape[-1])
```
Landanjs, line 164: do we need this? Or is just to be safe?

diffusion/datasets/laion/laion.py
```diff
@@ -97,9 +97,10 @@ def __getitem__(self, index):
             padding='max_length',
             max_length=self.tokenizer.model_max_length,
             truncation=True,
-        )['input_ids']
-        tokenized_caption = torch.tensor(tokenized_caption)
-        out = {'image': img, 'captions': tokenized_caption}
+        )
+        input_ids = torch.tensor(tokenized_caption['input_ids'])
+        attention_mask = torch.tensor(tokenized_caption['attention_mask'])
```
Landanjs, line 102: Should just set `return_tensors='pt'`?

mvpatel2000 (commented):
looks reasonable to me but I would want a manual test first

PR #20, opened 2023-05-24
Refresh Mosaicml platform yamls

aspfohl:
Similar changes as https://github.com/mosaicml/examples/pull/335

### DIFF
--- a/scripts/laion2b-en-interactive.yaml
+++ b/scripts/laion2b-en-interactive.yaml
@@ -1,8 +1,12 @@
-run_name: download-laion2b-en
-platform: {YOUR CLUSTER HERE}
-gpu_num:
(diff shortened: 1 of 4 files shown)


aspfohl:
@mvpatel2000 The changes to the yaml schema are all backwards compatible, but would like the reference yamls to use the new `compute` field for consistency: https://docs.mosaicml.com/projects/mcli/en/latest/training/yaml_schema.html#compute-fields

Specifically, there's a lot of new ways to request resources for your run, not just number of gpus. 

I also would like cluster & gpu_type to be commented out whenever possible, because it reduces friction for anyone who is getting started and only has a single gpu instance. All of these can be overridden by the CLI too, so makes it easier to move across multiple clusters without having to open and update a file

PR #23, opened 2023-05-30
Update Slack link in README.md

ejyuen:
Currently changing the public facing community slack links to : https://mosaicml.me/slack

We used a URL shortener so that if the Slack link dies on us we don't have to change all the slack links everywhere.

Please check if this works (in incognito)


Landanjs (approved):
LGTM

PR #40, opened 2023-06-20
Generic image-caption dataset

Landanjs:
The plan is to use this for both COCO and LAION datasets in the future and to remove these specific datasets in the future.

This also separates the logic to process precomputed latents into another class. PR for this coming soon

### DIFF
--- a/diffusion/datasets/image_caption.py
+++ b/diffusion/datasets/image_caption.py
@@ -1,0 +1,208 @@
+# Copyright 2022 MosaicML Diffusion authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Streaming Image-Caption dataset."""
+
+import random
+from io import BytesIO
+from typing import Callable, List, Optional, Sequence, Union
+
+import torch
+from PIL import Image
+from streaming import Stream, StreamingDataset
+from torch.utils.data import DataLoader
+from torchvision import transforms
+from transformers import AutoTokenizer
+
+from diffusion.datasets.laion.transforms import LargestCenterSquare
+
+# Disable PIL max image size limit
+Image.MAX_IMAGE_PIXELS = None
+
+
+class StreamingImageCaptionDataset(StreamingDataset):
+    """Streaming dataset for image-caption pairs.
+
+    Args:
+        streams (Sequence[Stream], optional): One or more Streams to stream/cache samples from.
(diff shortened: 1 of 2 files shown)


diffusion/datasets/image_caption.py
```diff
@@ -0,0 +1,208 @@
...
+        super().__init__(
+            streams=streams,
+            remote=remote,
+            local=local,
+            split=split,
+            shuffle=shuffle,
+            **streaming_kwargs,
+        )
+
+        if caption_selection not in ['first', 'random']:
```
coryMosaicML, line 70: Probably should do the `.lower()` here before the check

diffusion/datasets/image_caption.py
```diff
@@ -0,0 +1,208 @@
...
+        image_key=image_key,
+        caption_key=caption_key,
+        predownload=predownload,
+        download_retry=download_retry,
+        download_timeout=download_timeout,
+        batch_size=batch_size,
+        num_canonical_nodes=num_canonical_nodes,
+    )
+    # Create a subset of the dataset
+    if num_samples is not None:
```
coryMosaicML, line 197: We should probably get rid of this. I don't even know if it works...

diffusion/datasets/image_caption.py
```diff
@@ -0,0 +1,208 @@
...
+    # Setup the transforms to apply
+    if transform is None:
+        transform = [
+            LargestCenterSquare(resize_size),
+            transforms.ToTensor(),
+            transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5)),  # # Normalize from 0 to 1 to -1 to 1
+        ]
+    transform = transforms.Compose(transform)
+
+    dataset = StreamingImageCaptionDataset(
```
coryMosaicML, line 179: We should add an option to set `cache_limit` for the dataset.

coryMosaicML (approved):
Looks great! Much nicer than before.

diffusion/datasets/image_caption.py
Landanjs: 🤦

diffusion/datasets/image_caption.py
Landanjs: Should be covered by the new `streaming_kwargs`

PR #49, opened 2023-07-03
Add option to set the prediction type for stable diffusion

coryMosaicML:
This PR adds the option to set different prediction types following the hugging face naming convention ('epsilon' for noise prediction 'sample' for sample prediction, and 'v_prediction' for velocity prediction). It was something of an oversight that these were not already included. 

[Test run with v-prediction](https://wandb.ai/mosaic-ml/stable-diffusion-256-subsets/runs/crx8ifgv)

### DIFF
--- a/diffusion/models/stable_diffusion.py
+++ b/diffusion/models/stable_diffusion.py
@@ -186,7 +186,7 @@ class StableDiffusion(ComposerModel):
         elif self.prediction_type == 'sample':
             targets = latents
         elif self.prediction_type == 'v_prediction':
-            targets = self.scheduler.get_velocity(latents, noise, timesteps)
+            targets = self.noise_scheduler.get_velocity(latents, noise, timesteps)
         else:
             raise ValueError(
                 f'prediction type must be one of sample, epsilon, or v_prediction. Got {self.prediction_type}')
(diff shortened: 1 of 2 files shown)


diffusion/models/stable_diffusion.py
```diff
@@ -91,6 +90,9 @@ def __init__(self,
         self.vae = vae
         self.noise_scheduler = noise_scheduler
         self.loss_fn = loss_fn
+        if prediction_type.lower() not in ['sample', 'epsilon', 'v_prediction']:
+            raise ValueError(f'prediction type must be one of sample, epsilon, or v_prediction. Got {prediction_type}')
+        self.prediction_type = prediction_type.lower()
```
mvpatel2000, line 95:
```suggestion        
        self.prediction_type = prediction_type.lower()
        if prediction_type not in ['sample', 'epsilon', 'v_prediction']:
            raise ValueError(f'prediction type must be one of sample, epsilon, or v_prediction. Got {prediction_type}')
```
nit a little cleaner to move lower to top and do once

mvpatel2000 (approved):

diffusion/models/stable_diffusion.py
coryMosaicML: Done!

PR #51, opened 2023-07-18
Add evaluation using clean-fid

coryMosaicML:
This PR adds a script to configure and run evaluation through [clean-fid ](https://github.com/GaParmar/clean-fid).

### DIFF
--- a/diffusion/evaluation/clean_fid_eval.py
+++ b/diffusion/evaluation/clean_fid_eval.py
@@ -1,0 +1,236 @@
+# Copyright 2022 MosaicML Diffusion authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Evaluation using the clean-fid package."""
+
+import json
+import os
+from typing import List, Optional
+
+import clip
+import torch
+import wandb
+from cleanfid import fid
+from composer import ComposerModel, Trainer
+from composer.core import get_precision_context
+from composer.loggers import LoggerDestination, WandBLogger
+from composer.utils import dist
+from torch.utils.data import DataLoader
+from torchmetrics.multimodal import CLIPScore
+from torchvision.transforms.functional import to_pil_image
+from tqdm.auto import tqdm
+from transformers import PreTrainedTokenizerBase
+
+os.environ['TOKENIZERS_PARALLELISM'] = 'false'
+
+
+class CleanFIDEvaluator:
+    """Evaluator for CLIP, FID, KID, CLIP-FID scores using clean-fid.
+
+    See https://github.com/GaParmar/clean-fid for more information on clean-fid.
+
+    CLIP scores are computed using the torchmetrics CLIPScore metric.
+
+    Args:
+        model (ComposerModel): The model to evaluate.
+        eval_dataloader (DataLoader): The dataloader to use for evaluation.
+        clip_metric (CLIPScore): The CLIPScore metric to use for evaluation.
+        load_path (str, optional): The path to load the model from. Default: ``None``.
+        guidance_scales (List[float]): The guidance scales to use for evaluation.
+            Default: ``[1.0]``.
+        size (int): The size of the images to generate. Default: ``256``.
+        batch_size (int): The per-device batch size to use for evaluation. Default: ``16``.
+        loggers (List[LoggerDestination], optional): The loggers to use for logging results. Default: ``None``.
+        seed (int): The seed to use for evaluation. Default: ``17``.
+        output_dir (str): The directory to save results to. Default: ``/tmp/``.
+        num_samples (int, optional): The maximum number of samples to generate. Depending on batch size, actual
+            number may be slightly higher. If not specified, all the samples in the dataloader will be used.
+            Default: ``None``.
+        precision (str): The precision to use for evaluation. Default: ``'amp_fp16'``.
+        prompts (List[str], optional): The prompts to use for image visualtization.
+            Default: ``["A shiba inu wearing a blue sweater]``.
+
+    """
+
+    def __init__(self,
+                 model: ComposerModel,
+                 eval_dataloader: DataLoader,
+                 clip_metric: CLIPScore,
+                 load_path: Optional[str] = None,
+                 guidance_scales: Optional[List[float]] = None,
+                 size: int = 256,
+                 batch_size: int = 16,
+                 loggers: Optional[List[LoggerDestination]] = None,
+                 seed: int = 17,
+                 output_dir: str = '/tmp/',
+                 num_samples: Optional[int] = None,
+                 precision: str = 'amp_fp16',
+                 prompts: Optional[List[str]] = None):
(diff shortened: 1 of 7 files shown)


Landanjs:
Could you also add an example yaml to run eval?

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -0,0 +1,237 @@
...
+                load_path=self.load_path,
+                load_weights_only=True,
+                eval_dataloader=self.eval_dataloader)
+
+        # Move CLIP metric to device
+        self.device = dist.get_local_rank()
+        self.clip_metric = self.clip_metric.to(self.device)
+
+        # Predownload the CLIP model for computing clip-fid
+        _, _ = clip.load('ViT-B/32', device=self.device)
```
Landanjs, line 101: Is this required?

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -0,0 +1,237 @@
...
+        prompts = {}
+        # Iterate over the eval dataloader
+        num_batches = len(self.eval_dataloader)
+        for batch_id, batch in tqdm(enumerate(self.eval_dataloader)):
+            # Break if enough samples have been generated
+            if batch_id * self.batch_size * dist.get_world_size() >= self.num_samples:
+                break
+
+            real_images = batch['image']
+            captions = batch['captions']
```
Landanjs, line 130: Could 'image' and 'captions' be `image_key` and `caption_key` arguments in the `__init__`? I'm hesitant to add more arguments to the init, but might be a good precaution?

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -0,0 +1,237 @@
...
+            # Ensure a new seed for each batch, as randomness in model.generate is fixed.
+            starting_seed = self.seed + num_batches * dist.get_local_rank()
+            seed = starting_seed + batch_id
```
Landanjs, line 133: Nit: What's the reason for `starting_seed`? Would this be the same behavior?
```suggestion
            starting_seed = self.seed + batch_id * dist.get_local_rank()
```

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -0,0 +1,237 @@
...
+            # Generate images from the captions
+            with get_precision_context(self.precision):
+                generated_images = self.model.generate(tokenized_prompts=captions,
+                                                       height=self.size,
+                                                       width=self.size,
+                                                       guidance_scale=guidance_scale,
+                                                       seed=seed,
+                                                       progress_bar=False)  # type: ignore
+            # Get the prompts from the tokens
+            text_captions = [self.tokenizer.decode(caption, skip_special_tokens=True) for caption in captions]
```
Landanjs, line 143: Nit: This should work, right?
```suggestion
            text_captions = self.tokenizer.batch_decode(captions, skip_special_tokens=True)
```

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -0,0 +1,237 @@
...
+            self.clip_metric.update((generated_images * 255).to(torch.uint8), text_captions)
+            # Save the real images
+            for i, img in enumerate(real_images):
+                # Real images are -1, 1, so rescale to 0, 1.
+                to_pil_image((img + 1) / 2).save(f'{real_image_path}/{batch_id}_{i}_rank_{dist.get_local_rank()}.png')
```
Landanjs, line 148: The real images will only be [-1, 1] if the images are normalized, do we want to make this assumption? It seems more natural to assume unnormalized images which can be done by specifying the appropriate transforms in `build_streaming_image_caption_dataloader()`

Landanjs (approved):
Looks great, very detailed and structured! A few nits, a question about image normalization and wondering if we need the extra CLIP dependency.

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: This isn't strictly required, but it helps to do this download up front. Otherwise the clip model gets downloaded the first time the clip-fid metric is computed. This adds an extra import here, but not an extra dependency since the `clip` package is already required to compute `clean-fid`'s clip-fid metric.

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: `starting_seed` exists here to ensure that every image is generated using different randomness. This isn't the only way to ensure that though. If I understand correctly, the change you proposed here won't have the same effect, because different ranks will see the same seed, and hence generate the same random sequence. Example:

Current behavior:
```
seed = 17
num_batches = 3

batch_id   rank 0 seed   rank 1 seed
0            17               20                
1            18               21
2            19               22
```

Proposed behavior (if I understand the suggestion correctly):
```
seed = 17
num_batches = 3

batch_id   rank 0 seed   rank 1 seed
0            17               17                
1            17               21
2            17               22
```

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: We'll have to make an assumption here, or allow it to be configurable. Unnormalized images could mean multiple things depending on the specific dataset in question. Likely they are either in [0, 255] or [0, 1]. Should it be configurable? If not, what do you think the assumption should be?

diffusion/evaluation/clean_fid_eval.py
Landanjs: If the clip package is installed with `clean-fid`, do we need to include it in the setup.py? i.e. `clip@git+https://github.com/openai/CLIP.git`

diffusion/evaluation/clean_fid_eval.py
Landanjs: Ah, I needed `batch_id + 1` for it to work.

diffusion/evaluation/clean_fid_eval.py
Landanjs: I'd rather make an assumption than make it configurable. 

The current assumption is that the images will be [-1, 1] due to normalization, right? I think [0, 1] would be better assumption. What do you think?

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: Done!

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: Yes, it works too. Changed to this.

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: Ok, switched to default of [0, 1] and raise a warning if images are not in this range.

PR #52, opened 2023-07-18
Update inference script and add checkpoint path argument

Landanjs:
There have been a lot of breaking changes to inference since the stable diffusion launch. This PR gets inference working again and adds the ability to specify different checkpoints for inference.


mvpatel2000 (approved):
Assume u manually tested

PR #55, opened 2023-08-09
Update dependency versions for running on H100s and sidestep dataloader crash

coryMosaicML:
This PR updates dependency versions to enable running on H100s. Warning: this is backwards incompatible with checkpoints saved with earlier versions due to a change in the `transformers` library.

Additionally, `time.sleep()`s have been added after dataloader init to avoid an issue that caused a stochastic failure when initializing the streaming dataloaders. This is a temporary workaround until the fix is upstreamed to `streaming`.

### DIFF
--- a/diffusion/train.py
+++ b/diffusion/train.py
@@ -4,6 +4,7 @@
 """Train model."""
 
 import operator
+import time
 from collections.abc import Iterable
 from typing import Any, Dict, List, Optional, Union
 
@@ -38,6 +39,8 @@ def train(config: DictConfig) -> None:
         config.dataset.train_dataset,
         batch_size=config.dataset.train_batch_size // dist.get_world_size(),
     )
+    # Need to sleep for a bit to avoid dataloader crash
+    time.sleep(10)
 
     # Composer can take dataloaders, dataspecs, evaluators, or list of evaluators
     eval_set: Optional[Union[DataSpec, List[Evaluator]]] = None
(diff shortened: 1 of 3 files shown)


diffusion/train.py
```diff
@@ -38,6 +39,8 @@ def train(config: DictConfig) -> None:
         config.dataset.train_dataset,
         batch_size=config.dataset.train_batch_size // dist.get_world_size(),
     )
+    # Need to sleep for a bit to avoid dataloader crash
+    time.sleep(10)
```
Landanjs, line 43: What's the reason to not use `dist.barrier`?

setup.py
```diff
@@ -6,16 +6,16 @@
 from setuptools import find_packages, setup
 
 install_requires = [
-    'mosaicml@git+https://github.com/mosaicml/composer.git@6cf3d3a1aa300834c650f89460b5ac9bbc5a1e46',
+    'mosaicml',
```
Landanjs, line 9: Should we pin this?

diffusion/train.py
coryMosaicML: This is just a hacky workaround until the upstream issue is fixed. I went with a `sleep` rather than a barrier after discussing with @knighton who suggested that this might be due to the shared memory state init not being done, and streaming's dataset has a barrier near the end of its init already.  In my testing adding a barrier seems to also work, but one has to handle calling `initialize_dist` within `train.py` to do it which winds up being kind of strange.

setup.py
coryMosaicML: Done.

diffusion/train.py
Landanjs: Thanks for the explanation!

Landanjs (approved):
LGTM!

PR #61, opened 2023-08-17
Add partial SDXL model

jazcollins:
This PR adds `diffusion.models.models.stable_diffusion_xl`, which incorporates a few ideas from SDXL. Namely:

- Training with a larger UNet
- Loading the pre-trained SDXL VAE

Currently we're using the `madebyollin/sdxl-vae-fp16-fix` VAE checkpoint rather than the official `stabilityai/stable-diffusion-xl-base-1.0` VAE checkpoint. The stabilityai checkpoint leads to NaNs when training with fp16. More on that issue [here](https://huggingface.co/stabilityai/sdxl-vae/discussions/6).


diffusion/models/models.py
```diff
@@ -123,6 +123,122 @@ def stable_diffusion_2(
...
+    Requires batches of matched images and text prompts to train. Generates images from text
+    prompts. Currently uses UNet and VAE config from SDXL, but text encoder/tokenizer from SD2.
+
+    Args:
+        model_name (str, optional): Name of the model to load. Determines the text encoder, tokenizer,
+            and noise scheduler. Defaults to 'stabilityai/stable-diffusion-2-base'.
+        unet_model_name (str, optional): Name of the UNet model to load. Defaults to
+            'stabilityai/stable-diffusion-xl-base-1.0'.
+        vae_model_name (str, optional): Name of the VAE model to load. Defaults to
+            'madebyollin/sdxl-vae-fp16-fix'.
```
coryMosaicML, line 152: It might help to make a note of the fp16 issue here that makes this the necessary default.

diffusion/models/models.py
```diff
@@ -123,6 +123,122 @@ def stable_diffusion_2(
     return model
 
 
+def stable_diffusion_xl(
+    model_name: str = 'stabilityai/stable-diffusion-2-base',
```
coryMosaicML, line 127: Do we want to default this to use the stable diffusion XL tokenizers and encoder(s), or does that not work until we have the full implementation?

coryMosaicML (approved):
Looks great! Only a minor comment.

diffusion/models/models.py
```diff
@@ -123,6 +123,122 @@ def stable_diffusion_2(
...
+        val_guidance_scales = [1.0, 3.0, 7.0]
+    if loss_bins is None:
+        loss_bins = [(0, 1)]
+    # Fix a bug where CLIPScore requires grad
+    for metric in val_metrics:
+        if isinstance(metric, CLIPScore):
+            metric.requires_grad_(False)
+
+    if pretrained:
+        unet = UNet2DConditionModel.from_pretrained(unet_model_name, subfolder='unet')
```
Landanjs, line 183: I should I thought of this... loading pre-trained weights won't work since we don't have the full model setup, particularly the text encoder. We should get this working in the near future though

For now, I'd say throw a `NotImplementedError` and specify this in the docstring. Once the additional conditioning is added, we can add the additional encoder and any other components

diffusion/models/models.py
```diff
@@ -123,6 +123,122 @@ def stable_diffusion_2(
     return model
 
 
+def stable_diffusion_xl(
+    model_name: str = 'stabilityai/stable-diffusion-2-base',
+    unet_model_name: str = 'stabilityai/stable-diffusion-xl-base-1.0',
+    vae_model_name: str = 'madebyollin/sdxl-vae-fp16-fix',
```
Landanjs, line 129: Could we check if the SDXL VAE NaNs with bfloat16 within a few iteration? But there does look like weird colorization in the SDXL 1.0 VAE, so should we prefer this model?

Landanjs (commented):
Looks great! The only major adjustment is the `NotImplementedError` for the `pretrained` arg. I also had a question around which VAE we should prefer if we can run it on bf16.

For the future, there should be a future PR that adds the other modeling features (in addition to the conditioning PR), so we can run a pre-trained SDXL. Sorry for going back-and-forth on whether to fully implement SDXL

diffusion/models/models.py
jazcollins: Added!

diffusion/models/models.py
jazcollins: Waiting until the full implementation to do that, since there are two text encoders for SDXL.

diffusion/models/models.py
```diff
@@ -123,6 +123,122 @@ def stable_diffusion_2(
...
+    encode_latents_in_fp16: bool = True,
+    fsdp: bool = True,
+):
+    """Stable diffusion 2 training setup + SDXL UNet and VAE.
+
+    Requires batches of matched images and text prompts to train. Generates images from text
+    prompts. Currently uses UNet and VAE config from SDXL, but text encoder/tokenizer from SD2.
+
+    Args:
+        model_name (str, optional): Name of the model to load. Determines the text encoder, tokenizer,
```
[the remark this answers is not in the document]
jazcollins: Fixed this in my latest commit! Also, fixed it for `stable_diffusion_2`.

diffusion/models/models.py
jazcollins: Done.

diffusion/models/models.py
jazcollins: Will default to the `sdxl-vae-fp16-fix` VAE checkpoint for now, but will check about using the `stable-diffusion-xl-base-1.0` VAE checkpoint with bfloat16.

PR #64, opened 2023-09-06
Update xformers pin

Skylion007:
Updates xformers pin to 0.21 as it is the only version that has the right binaries needed to run it on H100s.


Landanjs (approved):
Thanks!

PR #71, opened 2023-10-05
support SDXL inference & update SD2/SDXL inference docstrings

jazcollins:
- Add StableDiffusionXLInference() class to inference
- Update docstrings for StableDiffusionInference (SD2) and StableDiffusionXLInference


Landanjs (approved):
LGTM!

PR #72, opened 2023-10-06
Add rescaled classifier free guidance and offset noise

coryMosaicML:
This PR adds two simple tricks that are claimed to improve image quality.

Trick 1: Rescaled classifier free guidance (see sec. 3.4 [here](https://arxiv.org/abs/2305.08891))
- A simple rescaling of the latents during classifier free guidance to help ameliorate the image over exposure problem.

Trick 2: Offset noise (see [here](https://www.crosslabs.org/blog/diffusion-with-offset-noise))
- A trick for using to fine tune models such that they can generate lighter/darker images. Adds a slight offset to the noise during training.

### DIFF
--- a/diffusion/models/stable_diffusion.py
+++ b/diffusion/models/stable_diffusion.py
@@ -39,6 +39,8 @@ class StableDiffusion(ComposerModel):
         loss_fn (torch.nn.Module): torch loss function. Default: `F.mse_loss`.
         prediction_type (str): The type of prediction to use. Must be one of 'sample',
             'epsilon', or 'v_prediction'. Default: `epsilon`.
+        offset_noise (bool): Whether or not to use offset noise. Default `False`.
+        offset_noise_scale (float): The scale of the offset noise. Default `0.1`.
         train_metrics (list): List of torchmetrics to calculate during training.
             Default: `None`.
         val_metrics (list): List of torchmetrics to calculate during validation.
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

diffusion/models/stable_diffusion.py
```diff
@@ -332,7 +341,9 @@ def generate(
         width: Optional[int] = None,
         num_inference_steps: Optional[int] = 50,
         guidance_scale: Optional[float] = 3.0,
-        num_images_per_prompt: int = 1,
+        rescale_guidance: bool = False,
+        rescaled_guidance_scale: float = 0.7,
```
jazcollins, line 345: If you wanted to be even more concise, instead of having both the `rescale_guidance: bool = False` and `rescaled_guidance_scale: float = 0.7` parameters, you could just pass around a single `rescale_guidance: Optional[**float**] = None`. If it is None, don't apply rescale guidance - if it is set to something (e.g. 0.7), apply rescale guidance with that value.
 
 Later on in generate when you want to do the rescaling, you'd have:
 
```
# Optionally rescale the classifer free guidance
if rescale_guidance:
    std_pos = torch.std(pred_text, dim=(1, 2, 3), keepdim=True)
    std_cfg = torch.std(pred, dim=(1, 2, 3), keepdim=True)
    pred_rescaled = pred * (std_pos / std_cfg)
    pred = pred_rescaled * rescale_guidance + pred * (1 - rescale_guidance)
```

jazcollins (approved):
Looks good to me! 

If you wanted to make things even more concise, I left one suggestion to consolidate the boolean and float parameters for rescale guidance into a single Optional[float] parameter. The same could be done for the offset noise parameters - but what you wrote looks great and it is not necessary!

Landanjs (approved):
LGTM. +1 to Jasmine's suggestion

diffusion/models/stable_diffusion.py
Skylion007: ^ Great idea.

coryMosaicML: Thanks for the suggestion! Should be that way now.

PR #73, opened 2023-10-06
Adjusts the synthetic dataset to work with SDXL

Landanjs:
SDXL requires additional inputs from the dataset. This adds the additional inputs to the synthetic dataset class, so we use the synthetic dataset with SDXL


jazcollins (approved):
Looks good to me, thanks!

PR #74, opened 2023-10-06
Landan/aspect ratio bucketing

Landanjs:
This PR adds a new crop transformation that enables using aspect ratio buckets. `RandomCropAspectRatioTransform` has a predefined set of aspect ratio buckets based on the SDXL paper. The transform determines which bucket an image belongs, then resizes and crops the image to fit in the designated bucket.

Since this adds another cropping option, I swapped `random_crop` for `crop_type` which can select from 'square', 'random', and 'aspect_ratio'.

I also made the `LargestCenterSquare` transform return cropping parameters to match the functionality of 'square' and 'random' crop options. 

I did some refactoring as I was adding this, let me know if it wasn't appropriate!

### DIFF
--- a/diffusion/datasets/image_caption.py
+++ b/diffusion/datasets/image_caption.py
@@ -15,7 +15,7 @@ from torch.utils.data import DataLoader
 from torchvision import transforms
 from transformers import AutoTokenizer
 
-from diffusion.datasets.laion.transforms import RandomCropSquareReturnTransform, LargestCenterSquareReturnTransform, RandomCropAspectRatioTransorm
+from diffusion.datasets.laion.transforms import LargestCenterSquare, RandomCropSquareReturnTransform, RandomCropAspectRatioTransorm
 from diffusion.models.models import SDXLTokenizer
 
 log = logging.getLogger(__name__)
@@ -132,7 +132,6 @@ class StreamingImageCaptionDataset(StreamingDataset):
             if torch.rand(1) < self.microcond_drop_prob:
                 out['cond_target_size'] = out['cond_target_size'] * 0
 
-
         # Caption
         if torch.rand(1) < self.caption_drop_prob:
             caption = ''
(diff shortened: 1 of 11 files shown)


diffusion/datasets/image_caption.py
```diff
@@ -73,38 +72,54 @@ def __init__(
...
+            img = self.transform(img)
+        out['image'] = img
+
+        # SDXL microconditioning on image characteristics
+        if self.sdxl:
+            # Get the new height and width
+            if isinstance(img, torch.Tensor):
+                img_h, img_w = img.shape[-2], img.shape[-1]
+            elif isinstance(img, Image.Image):
+                img_h, img_w = img.size
```
jazcollins, line 116: `img_h` and `img_w` should also be swapped, same thing as above

diffusion/datasets/image_caption.py
```diff
@@ -73,38 +72,54 @@ def __init__(
...
         sample = super().__getitem__(index)
+        out = {}
 
         # Image
         img = sample[self.image_key]
         if not isinstance(img, Image.Image):
             img = Image.open(BytesIO(sample[self.image_key]))
         if img.mode != 'RGB':
             img = img.convert('RGB')
+        orig_h, orig_w = img.size
```
jazcollins, line 99: According to PIL Image [docs](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.size), `img.size` will return (width, height) tuple, so `orig_h` and `orig_w` should be flipped.

jazcollins (commented):
Overall looks great! Plus, the way you separated out the crop transform from the rest of the transforms is a lot cleaner than what we were doing before.

Only thing that stuck out to me was two instances of mixed up height/width - see the comments

diffusion/datasets/image_caption.py
```diff
@@ -57,8 +56,8 @@ def __init__(
         caption_drop_prob: float = 0.0,
         microcond_drop_prob: float = 0.0,
         caption_selection: str = 'first',
+        crop: Optional[Callable] = None,
```
jazcollins, line 59: Add blurb about this in docs

`crop_type (str, optional): Type of crop to perform, either ['square', 'random', 'aspect_ratio']. Default: ``'square'``.`

jazcollins (approved):
LGTM

PR #77, opened 2023-10-09
Zero dropped captions

jazcollins:
Previously when we applied caption dropout (via `caption_drop_prob` in the dataloader), those dropped captions (i.e. empty strings) would end up getting tokenized into padding tokens and encoded by the text encoder(s). This PR adds the option to instead encode the dropped captions as an all zeros tensor by setting `zero_dropped_caption=True`. Currently this flag is set such that zero-ing out dropped captions will become the new default behavior for both SD2 and SDXL training.

### DIFF
--- a/diffusion/datasets/image_caption.py
+++ b/diffusion/datasets/image_caption.py
@@ -43,6 +43,8 @@ class StreamingImageCaptionDataset(StreamingDataset):
         image_key (str): Key associated with the image in the streaming dataset. Default: ``'image'``.
         caption_key (str): Key associated with the caption in the streaming dataset. Default: ``'caption'``.
         sdxl (bool): Whether or not we're training SDXL. Default: `False`.
+        zero_dropped_captions (bool): If True, zero out text embeddings for dropped captions. Default: ``True``.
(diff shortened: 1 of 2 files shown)


coryMosaicML (approved):
Looks good!

diffusion/datasets/image_caption.py
```diff
@@ -43,6 +43,8 @@ class StreamingImageCaptionDataset(StreamingDataset):
         image_key (str): Key associated with the image in the streaming dataset. Default: ``'image'``.
         caption_key (str): Key associated with the caption in the streaming dataset. Default: ``'caption'``.
         sdxl (bool): Whether or not we're training SDXL. Default: `False`.
+        zero_dropped_captions (bool): If True, zero out text embeddings for dropped captions. Default: ``True``.
+
```
Landanjs, line 47:
```suggestion
        zero_dropped_captions (bool): If True, zero out text embeddings for dropped captions. Default: ``True``.
```

Landanjs (approved):
LGTM!

PR #79, opened 2023-10-11
Autoencoder implementation and training

coryMosaicML:
This PR adds an implementation of the latent diffusion autoencoder, loss function, and training setup for training the autoencoder in composer. 

- Basic network blocks are in `diffusion/models/layers.py`
- Complete networks, loss, and the and composer models are in `diffusion/models/autoencoder.py`
- For convenience in visualizing image reconstructions, there's a new image logging callback in `diffusion/callbacks/log_diffusion_images.py`
- `diffusion/algorithms/discriminator_schedule.py` contains a composer algorithm for switching on the discriminator.
- `diffusion/train.py` also contains some modifications to configure the param groups for training the autoencoder.


diffusion/algorithms/discriminator_schedule.py
```diff
@@ -0,0 +1,61 @@
...
+        return event in [Event.INIT, Event.BATCH_START, Event.AFTER_LOSS]
+
+    def apply(self, event: Event, state: State, logger: Logger) -> None:
+        if hasattr(state.model, 'autoencoder_loss') and isinstance(state.model.autoencoder_loss, nn.Module):
+            autoencoder_loss = state.model.autoencoder_loss
+        elif isinstance(state.model.module, nn.Module) and hasattr(state.model.module, 'autoencoder_loss'):
+            autoencoder_loss = state.model.module.autoencoder_loss
+        else:
+            raise ValueError('Model does not have a discriminator weight')
+        assert isinstance(autoencoder_loss, nn.Module) and callable(autoencoder_loss.set_discriminator_weight)
```
Skylion007, line 40:
```suggestion
        assert isinstance(autoencoder_loss, nn.Module) and callable(autoencoder_loss.set_discriminator_weight), f"{type(autoencoder_loss)}"

File: diffusion/models/autoencoder.py [hunk 2]
```diff
@@ -0,0 +1,613 @@
...
+        clamped_imgs = outputs['x_recon'].clamp(-1, 1)
+        if isinstance(metric, MeanMetric):
+            metric.update(torch.square(outputs['latents']))
+        elif isinstance(metric, LearnedPerceptualImagePatchSimilarity):
+            metric.update(clamped_imgs, batch[self.input_key])
+        elif isinstance(metric, PeakSignalNoiseRatio):
+            metric.update(clamped_imgs, batch[self.input_key])
+        elif isinstance(metric, StructuralSimilarityIndexMeasure):
+            metric.update(clamped_imgs, batch[self.input_key])
+        else:
```
Skylion007, line 585: Do you want to add an isintance assert about what the metric is suppose to be here?

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+        """Forward through the attention layer."""
+        # Need to remember H, W to get back to it
+        H, W = x.shape[2:]
+        h = self.norm(x)
+        # Get q, k, and v
+        qkv = self.qkv_conv(h)
+        q, k, v = torch.split(qkv, self.input_channels, dim=1)
+        # Use torch's built in attention function
+        q, k, v = self._reshape_for_attention(q), self._reshape_for_attention(k), self._reshape_for_attention(v)
+        h = F.scaled_dot_product_attention(q, k, v, dropout_p=self.dropout_probability)
```
Skylion007, line 348: No opt in xformers?

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+    scaling_layer = GradientScalingLayer()
+    scaling_layer.register_full_backward_hook(scaling_layer.backward_hook)
+
+    And then set the scale via
+    scaling_layer.set_scale(scale)
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.scale = 1.0
```
Skylion007, line 421: You could reduce some of this boilerplat by having a dataclass.
```suggestion
        self.scale : float = 1.0
```

diffusion/models/models.py
```diff
@@ -263,6 +264,137 @@ def stable_diffusion_xl(
...
+                      resample_with_conv: bool = True,
+                      zero_init_last: bool = False,
+                      input_key: str = 'image',
+                      learn_log_var: bool = True,
+                      log_var_init: float = 0.0,
+                      kl_divergence_weight: float = 1.0,
+                      lpips_weight: float = 0.25,
+                      discriminator_weight: float = 0.5,
+                      discriminator_num_filters: int = 64,
+                      discriminator_num_layers: int = 3):
```
Skylion007, line 285:
```suggestion
                      discriminator_num_layers: int = 3) -> ComposerAutoEncoder:
```

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
 from omegaconf import DictConfig, OmegaConf
+from torch.optim import Optimizer
+
+from diffusion.models.autoencoder import ComposerAutoEncoder, ComposerDiffusersAutoEncoder
+
+
+def make_autoencoder_optimizer(config: DictConfig, model: ComposerModel) -> Optimizer:
+    """Configures the optimizer for use with an autoencoder + discriminator loss."""
+    print('Configuring opimizer for autoencoder+discriminator')
+    assert isinstance(model, ComposerAutoEncoder) or isinstance(model, ComposerDiffusersAutoEncoder)
```
Skylion007, line 28:
```suggestion
    assert isinstance(model,  (ComposerAutoEncoder, ComposerDiffusersAutoEncoder))
```

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
+
+    # Configure optimizer settings for the autoencoder
+    if hasattr(config, 'autoencoder_optimizer'):
+        autoencoder_param_dict = {k: v for k, v in config.autoencoder_optimizer.items()}
```
Skylion007, line 32:
```suggestion
        autoencoder_param_dict = dict(config.autoencoder_optimizer.items())
```

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
+    else:
+        autoencoder_param_dict = {k: v for k, v in config.optimizer.items()}
```
Skylion007, line 34:
```suggestion
        autoencoder_param_dict = dict(config.optimizer.items())
```

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
+
+    if model.autoencoder_loss.learn_log_var:
+        autoencoder_param_dict['params'] = chain(model.model.parameters(), [model.autoencoder_loss.log_var])
+    else:
+        autoencoder_param_dict['params'] = model.model.parameters()
+
+    # Configure optimizer settings for the discriminator
+    if hasattr(config, 'discriminator_optimizer'):
+        discriminator_param_dict = {k: v for k, v in config.discriminator_optimizer.items()}
```
Skylion007, line 43:
```suggestion
        discriminator_param_dict = dict(config.discriminator_optimizer.items())
```

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
+    else:
+        discriminator_param_dict = {k: v for k, v in config.optimizer.items()}
```
Skylion007, line 45:
```suggestion
        discriminator_param_dict = dict(config.optimizer.items())
```

tests/test_autoencoder.py
```diff
@@ -0,0 +1,119 @@
...
+@pytest.mark.parametrize('dropout_probability', [0.0, 0.1])
+def test_resnet_block(input_channels, output_channels, use_conv_shortcut, dropout_probability):
+    block = ResNetBlock(input_channels=input_channels,
+                        output_channels=output_channels,
+                        use_conv_shortcut=use_conv_shortcut,
+                        dropout_probability=dropout_probability,
+                        zero_init_last=True)
+    x = torch.randn(1, input_channels, 5, 5)
+    y = block(x)
+    assert y.shape == (1, output_channels, 5, 5)
```
Skylion007, line 25:
```suggestion
    assert y.shape == (1, output_channels, 5, 5), f"{y.shape}"
```

diffusion/algorithms/discriminator_schedule.py
```diff
@@ -0,0 +1,61 @@
...
+        self.lr = 0.0
+        self.weight_decay = 0.0
+        self.discriminator_weight = 0.0
+
+        # Verify the start iteration is in units of batches
+        if self.start_iteration.unit != TimeUnit.BATCH:
+            raise ValueError('start_iteration must be in units of batches')
+
+    def match(self, event: Event, state: State) -> bool:
+        return event in [Event.INIT, Event.BATCH_START, Event.AFTER_LOSS]
```
Skylion007, line 31:
```suggestion
        return event in (Event.INIT, Event.BATCH_START, Event.AFTER_LOSS)
```

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward through the attention layer."""
+        # Need to remember H, W to get back to it
+        H, W = x.shape[2:]
+        h = self.norm(x)
+        # Get q, k, and v
+        qkv = self.qkv_conv(h)
+        q, k, v = torch.split(qkv, self.input_channels, dim=1)
+        # Use torch's built in attention function
+        q, k, v = self._reshape_for_attention(q), self._reshape_for_attention(k), self._reshape_for_attention(v)
```
Landanjs, line 347: I'm not sure if there will be any performance difference, but I think you should reshape `qkv` then split it.

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+        """Reshape the input tensor from attention."""
+        # x is (B, H*W, C), need it to be (B, C, H, W) for conv
+        x = x.reshape(x.shape[0], H, W, -1).permute(0, 3, 1, 2).contiguous()
+        return x
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward through the attention layer."""
+        # Need to remember H, W to get back to it
+        H, W = x.shape[2:]
+        h = self.norm(x)
```
Landanjs, line 342: Is there usually a norm here?

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+        input_channels (int): Number of input channels.
+        resample_with_conv (bool): Whether to use a conv for downsampling.
+    """
+
+    def __init__(self, input_channels: int, resample_with_conv: bool):
+        super().__init__()
+        self.input_channels = input_channels
+        self.resample_with_conv = resample_with_conv
+        if self.resample_with_conv:
+            self.conv = nn.Conv2d(self.input_channels, self.input_channels, kernel_size=3, stride=2, padding=0)
```
Landanjs, line 369: Is there a reason this kernel has to be 3 instead of 2 or 4?

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+            x = block(x)
+        return x
+
+
+class AutoEncoderLoss(nn.Module):
+    """Loss function for training an autoencoder. Includes LPIPs and a discriminator.
+
+    Args:
+        input_key (str): Key for the input to the model. Default: `image`.
+        output_channels (int): Number of output channels. Default: `3`.
```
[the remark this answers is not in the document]
Landanjs: I suppose it's because the discriminator is taking the output from the AE, but it will also take the input as well? Eh not an important change...

diffusion/models/models.py
```diff
@@ -263,6 +264,137 @@ def stable_diffusion_xl(
...
 
 
+def build_autoencoder(input_channels: int = 3,
+                      output_channels: int = 3,
+                      hidden_channels: int = 128,
+                      latent_channels: int = 4,
+                      double_latent_channels: bool = True,
+                      channel_multipliers: Tuple[int, ...] = (1, 2, 4, 4),
+                      num_residual_blocks: int = 2,
+                      use_conv_shortcut=False,
```
Landanjs, line 274:
```suggestion
                      use_conv_shortcut: bool =False,
```

Landanjs (approved):
Looks good overall! The only bug I noticed was in the n layer discriminator and unlikely for us to hit it.

Most of my other comments are about where specific parameter settings came from. I'm sure for most of these you already thought about them and made a decision, so feel free to ignore my comments if they aren't important

diffusion/algorithms/discriminator_schedule.py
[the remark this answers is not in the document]
coryMosaicML: Vestigial from an earlier implementation. Thanks for catching!

diffusion/algorithms/discriminator_schedule.py
```diff
@@ -0,0 +1,61 @@
...
+            if state.timestamp.get(TimeUnit.BATCH).value >= self.start_iteration.value:
+                # Turn on the discriminator completely.
+                state.optimizers[0].param_groups[1]['lr'] = self.lr
+                state.optimizers[0].param_groups[1]['weight_decay'] = self.weight_decay
+                autoencoder_loss.set_discriminator_weight(self.discriminator_weight)
+            else:
+                # Turn off the discriminator
+                state.optimizers[0].param_groups[1]['lr'] = 0.0
+                state.optimizers[0].param_groups[1]['weight_decay'] = 0.0
+                autoencoder_loss.set_discriminator_weight(0.0)
```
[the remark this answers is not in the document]
coryMosaicML: Could be done on init since we never turn the discriminator back off again.

diffusion/algorithms/discriminator_schedule.py
coryMosaicML: Done!

diffusion/algorithms/discriminator_schedule.py
```diff
@@ -0,0 +1,61 @@
...
+        if self.start_iteration.unit != TimeUnit.BATCH:
+            raise ValueError('start_iteration must be in units of batches')
+
+    def match(self, event: Event, state: State) -> bool:
+        return event in [Event.INIT, Event.BATCH_START, Event.AFTER_LOSS]
+
+    def apply(self, event: Event, state: State, logger: Logger) -> None:
+        if hasattr(state.model, 'autoencoder_loss') and isinstance(state.model.autoencoder_loss, nn.Module):
+            autoencoder_loss = state.model.autoencoder_loss
+        elif isinstance(state.model.module, nn.Module) and hasattr(state.model.module, 'autoencoder_loss'):
```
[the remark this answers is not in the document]
coryMosaicML: You're correct, there would be an error if the model 1) does not contain an autoencoder loss, and 2) is not DDP wrapped. Case 1 would always cause an error, but case 1 with case 2 would cause the error to not be caught. I fixed it, which required refactoring these if-statements a bit.

diffusion/algorithms/discriminator_schedule.py
coryMosaicML: Done!

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+        self.output_channels = output_channels if output_channels is not None else input_channels
+        self.use_conv_shortcut = use_conv_shortcut
+        self.dropout_probability = dropout_probability
+        self.zero_init_last = zero_init_last
+
+        self.norm1 = nn.GroupNorm(num_groups=32, num_channels=self.input_channels, eps=1e-6, affine=True)
+        self.conv1 = nn.Conv2d(self.input_channels, self.output_channels, kernel_size=3, stride=1, padding=1)
+        nn.init.kaiming_normal_(self.conv1.weight, nonlinearity='linear')
+        # Output layer is immediately after a silu. Need to account for that in init.
+        self.conv1.weight.data *= 1.6761
```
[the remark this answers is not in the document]
coryMosaicML: It's a numerically computed value such that the output has variance 1.

diffusion/algorithms/discriminator_schedule.py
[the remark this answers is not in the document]
Landanjs: Definitely not a necessary change!

diffusion/algorithms/discriminator_schedule.py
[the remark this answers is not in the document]
Landanjs: The changes look great!

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        nn.init.kaiming_normal_(input_conv.weight, nonlinearity='linear')
+        self.blocks.append(input_conv)
+
+        out_filters = self.num_filters
+        for n in range(1, self.num_layers):
+            in_filters = self.num_filters * 2**(n - 1)
+            out_filters = self.num_filters * min(2**n, 8)
+            conv = nn.Conv2d(in_filters, out_filters, kernel_size=4, stride=2, padding=self.padding, bias=False)
+            # Init these as if a linear layer follows them because batchnorm happens before leaky relu.
+            nn.init.kaiming_normal_(conv.weight, nonlinearity='linear')
```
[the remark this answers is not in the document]
coryMosaicML: Yes, because of the batchnorm.

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+
+    def __init__(self, input_channels: int = 3, num_filters: int = 64, num_layers: int = 3, padding: int = 0):
+        super().__init__()
+        self.input_channels = input_channels
+        self.num_filters = num_filters
+        self.num_layers = num_layers
+        self.padding = padding
+
+        self.blocks = nn.ModuleList()
+        input_conv = nn.Conv2d(self.input_channels, self.num_filters, kernel_size=4, stride=2, padding=self.padding)
```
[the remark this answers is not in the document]
coryMosaicML: `padding=0` was leftover an old experiment and shouldn't be here. Removed it now. 

The missing leaky relu was an oversight, thanks for catching it! Added it now.

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        nn.init.kaiming_normal_(input_conv.weight, nonlinearity='linear')
+        self.blocks.append(input_conv)
+
+        out_filters = self.num_filters
+        for n in range(1, self.num_layers):
+            in_filters = self.num_filters * 2**(n - 1)
```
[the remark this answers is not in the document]
coryMosaicML: Good catch! Should be fixed now.

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        output_conv = nn.Conv2d(final_out_filters, 1, kernel_size=4, stride=1, padding=self.padding, bias=False)
+        nn.init.kaiming_normal_(output_conv.weight, nonlinearity='linear')
+        # Should init output layer such that outputs are generally within the linear region of a sigmoid.
+        output_conv.weight.data *= 0.1
+        self.blocks.append(output_conv)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward through the discriminator."""
+        for block in self.blocks:
+            x = block(x)
```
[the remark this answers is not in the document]
coryMosaicML: Switched it to a `nn.Sequential`

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        conv = nn.Conv2d(out_filters, final_out_filters, kernel_size=4, stride=1, padding=self.padding, bias=False)
+        nn.init.kaiming_normal_(conv.weight, nonlinearity='linear')
+        norm = nn.BatchNorm2d(final_out_filters)
+        nonlinearity = nn.LeakyReLU(0.2, True)
+        self.blocks.extend([conv, norm, nonlinearity])
+        # Output layer
+        output_conv = nn.Conv2d(final_out_filters, 1, kernel_size=4, stride=1, padding=self.padding, bias=False)
+        nn.init.kaiming_normal_(output_conv.weight, nonlinearity='linear')
+        # Should init output layer such that outputs are generally within the linear region of a sigmoid.
+        output_conv.weight.data *= 0.1
```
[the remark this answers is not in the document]
coryMosaicML: Yes, though in practice this makes little difference. Can remove it if it's unnecessary complexity.

diffusion/algorithms/discriminator_schedule.py
```diff
@@ -0,0 +1,63 @@
...
+    def apply(self, event: Event, state: State, logger: Logger) -> None:
+        if hasattr(state.model, 'autoencoder_loss'):
+            autoencoder_loss = state.model.autoencoder_loss
+        elif hasattr(state.model, 'module') and isinstance(state.model.module, nn.Module) and hasattr(
+                state.model.module, 'autoencoder_loss'):
+            autoencoder_loss = state.model.module.autoencoder_loss
+        else:
+            raise ValueError('Model does not have an autoencoder loss')
+        assert isinstance(autoencoder_loss, nn.Module) and callable(
+            autoencoder_loss.set_discriminator_weight), f'{type(autoencoder_loss)}'
```
Skylion007, line 42:
```suggestion
            autoencoder_loss.set_discriminator_weight), f'{type(autoencoder_loss)=}'
```
will print `type(autoencoder_loss)=VALUE`

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
coryMosaicML: Yes, it's because this value should be the autoencoder's `output_channels` value. Not sure which way would be less confusing here, although typically `output_channels = input_channels` anyways.

diffusion/algorithms/discriminator_schedule.py
coryMosaicML: Done!

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        # Ensure that lpips does not get trained
+        for param in self.lpips.parameters():
+            param.requires_grad_(False)
+        for param in self.lpips.net.parameters():
+            param.requires_grad_(False)
+
+        # Set up the discriminator
+        self.discriminator_num_filters = discriminator_num_filters
+        self.discriminator_num_layers = discriminator_num_layers
+        self.discriminator_weight = discriminator_weight
```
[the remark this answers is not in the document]
coryMosaicML: Fixed!

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        # LPIPs loss. Images for LPIPS must be in [-1, 1]
+        recon_img = outputs['x_recon'].clamp(-1, 1)
+        target_img = batch[self.input_key].clamp(-1, 1)
+        lpips_loss = self.lpips(recon_img, target_img)
+        losses['lpips_loss'] = lpips_loss.mean()
+
+        # Make the nll loss
+        rec_loss = ae_loss + self.lpips_weight * lpips_loss
+        # Note: the +2 here comes from the nll of the laplace distribution.
+        # It's only here to make you feel better by keeping the loss positive for longer.
```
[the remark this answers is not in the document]
coryMosaicML: The comment is probably not worthwhile, but the factor of 2 is helpful to avoid confusion with negative loss values during training. So I simplified the comment but left the factor of 2.

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        disc_weight, nll_grads_norm, disc_grads_norm = self.calc_discriminator_adaptive_weight(
+            nll_loss, fake_loss, last_layer)
+        losses['disc_weight'] = disc_weight
+        losses['nll_grads_norm'] = nll_grads_norm
+        losses['disc_grads_norm'] = disc_grads_norm
+
+        # Make the KL divergence loss (effectively regularize the latents)
+        mean = outputs['mean']
+        log_var = outputs['log_var']
+        kl_div_loss = -0.5 * (1 + log_var - mean.pow(2) - log_var.exp())
```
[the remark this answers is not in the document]
coryMosaicML: Can you show me how you're getting that? I think it is correct as is according to my derivation. I believe it also agrees with the implementation [here](https://github.com/CompVis/stable-diffusion/blob/21f890f9da3cfbeaba8e2ac3c425ee9e998d5229/ldm/modules/distributions/distributions.py#L44).

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        # Count the number of latent elements to normalize the loss
+        num_latent_elements = mean.numel() // kl_div_loss.shape[0]
+        losses['kl_div_loss'] = kl_div_loss.mean()
+
+        # Combine the losses. Downweight the kl_div_loss to account for differing dimensionalities.
+        dimensionality_weight = num_latent_elements / num_output_elements
```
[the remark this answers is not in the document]
coryMosaicML: This is to allow using a mean instead of a sum across pixels/latent dims to make the loss scale invariant to image resolution without changing the relative loss weights. It's a choice, I can roll it back if you like.

diffusion/models/autoencoder.py
coryMosaicML: What would we assert it be? A generic `torchmetrics.Metric` perhaps? The `else` here follows the [composer classifier](https://github.com/mosaicml/composer/blob/f65139c20da1476a610294ed5bb76f2a15bbae0d/composer/models/tasks/classification.py#L122) to avoid enumerating the specific types of metrics that might be used. Although there are cases where this could break if we make the metrics configurable (they aren't ATM).

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,613 @@
...
+        losses['disc_weight'] = disc_weight
+        losses['nll_grads_norm'] = nll_grads_norm
+        losses['disc_grads_norm'] = disc_grads_norm
+
+        # Make the KL divergence loss (effectively regularize the latents)
+        mean = outputs['mean']
+        log_var = outputs['log_var']
+        kl_div_loss = -0.5 * (1 + log_var - mean.pow(2) - log_var.exp())
+        # Count the number of latent elements to normalize the loss
+        num_latent_elements = mean.numel() // kl_div_loss.shape[0]
```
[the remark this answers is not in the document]
coryMosaicML: No, we don't need to round here. But there's a cleaner way to do this that avoids using division so I changed it.

diffusion/models/layers.py
coryMosaicML: I I understand the [docs](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) correctly, this should be enabled by default in newer PyTorch versions.

diffusion/models/layers.py
coryMosaicML: Agreed, made the change.

diffusion/models/layers.py
coryMosaicML: It's 3 from the original code [here](https://github.com/CompVis/stable-diffusion/blob/21f890f9da3cfbeaba8e2ac3c425ee9e998d5229/ldm/modules/diffusionmodules/model.py#L68). Are you asking about using 2 or 4 to avoid the weird padding?

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.scale = 1.0
+
+    def set_scale(self, scale: float):
+        self.scale = scale
+
+    def forward(self, x):
```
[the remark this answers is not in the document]
coryMosaicML: Can you elaborate?

diffusion/models/layers.py
coryMosaicML: Added the type here. Can you elaborate on the point about reducing the boilerplate?

diffusion/models/models.py
coryMosaicML: Fixed!

diffusion/train.py
coryMosaicML: Fixed, thanks!

diffusion/train.py
coryMosaicML: Fixed this, along with the other similar ones. Thanks!

tests/test_autoencoder.py
coryMosaicML: Added!

tests/test_autoencoder.py
```diff
@@ -0,0 +1,119 @@
...
+
+@pytest.mark.parametrize('height', [32])
+@pytest.mark.parametrize('width', [32])
+@pytest.mark.parametrize('num_layers', [3])
+def test_discriminator(height, width, num_layers):
+    discriminator = NlayerDiscriminator(input_channels=3, num_filters=16, num_layers=num_layers, padding=1)
+    x = torch.randn(1, 3, height, width)
+    y = discriminator(x)
+    downsample_factor = 2**(num_layers + 1)
+    assert y.shape == (1, 1, height // downsample_factor, width // downsample_factor)
```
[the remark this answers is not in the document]
coryMosaicML: Added1

tests/test_autoencoder.py
```diff
@@ -0,0 +1,119 @@
...
+                              double_latent_channels=True,
+                              channel_multipliers=(1, 2, 4, 4),
+                              num_residual_blocks=2,
+                              use_conv_shortcut=False,
+                              dropout_probability=0.0,
+                              resample_with_conv=True)
+    # Check that the number of parameters is the same
+    hf_params = sum(p.numel() for p in hf_autoencoder.parameters() if p.requires_grad)
+    params = sum(p.numel() for p in autoencoder.parameters() if p.requires_grad)
+    assert hf_params == params
```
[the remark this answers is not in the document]
coryMosaicML: Added!

diffusion/models/autoencoder.py
```diff
@@ -0,0 +1,607 @@
...
+    """
+
+    def __init__(self,
+                 input_channels: int = 3,
+                 hidden_channels: int = 128,
+                 latent_channels: int = 4,
+                 double_latent_channels: bool = True,
+                 channel_multipliers: Tuple[int, ...] = (1, 2, 4, 8),
+                 num_residual_blocks: int = 4,
+                 use_conv_shortcut=False,
```
Skylion007, line 44:
```suggestion
                 use_conv_shortcut : bool = False,
```

coryMosaicML: Fixed!

diffusion/models/autoencoder.py
Landanjs: If I understand, there is a fixed set of 5 metrics defined in the `__init__`. 4 of the metrics are caught by the above if-statement and the fifth one (`MeanSquaredError`, right?) is caught by this if-statement. It seems safer and consistent to check `elif isinstance(metric, MeanSquaredError)`

diffusion/models/layers.py
Skylion007: You are right.

diffusion/models/layers.py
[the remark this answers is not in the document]
Skylion007: Never mind, this comment is no longer relevant now that the this isntance checks were removed

diffusion/train.py
```diff
@@ -16,6 +17,37 @@
...
 from composer.utils import dist, reproducibility
 from omegaconf import DictConfig, OmegaConf
+from torch.optim import Optimizer
+
+from diffusion.models.autoencoder import ComposerAutoEncoder, ComposerDiffusersAutoEncoder
+
+
+def make_autoencoder_optimizer(config: DictConfig, model: ComposerModel) -> Optimizer:
+    """Configures the optimizer for use with an autoencoder + discriminator loss."""
+    print('Configuring opimizer for autoencoder+discriminator')
```
Skylion007, line 27: Wait, shouldn't this use a logger instead of a print statement?

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: Hmm, what if you add "this probably makes little difference" in the comment? It confused me when reading the code

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: Understandable

diffusion/models/layers.py
```diff
@@ -222,3 +223,208 @@ def __call__(
...
+    def __init__(self, input_channels: int, resample_with_conv: bool):
+        super().__init__()
+        self.input_channels = input_channels
+        self.resample_with_conv = resample_with_conv
+        if self.resample_with_conv:
+            self.conv = nn.Conv2d(self.input_channels, self.input_channels, kernel_size=3, stride=1, padding=1)
+            nn.init.kaiming_normal_(self.conv.weight, nonlinearity='linear')
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = F.interpolate(x, scale_factor=2, mode='nearest')
```
Skylion007, line 400: Do we need tow rry about align corners or antialias? Probably not, but worth considering.

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: How annoying would it be to set this to `ae_output_channels`?

diffusion/models/layers.py
Landanjs: Yes haha, but if that is how they do it 🤷

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: I was probably wrong, looks like it matches

diffusion/train.py
coryMosaicML: Probably, yeah. But the rest of `train.py` uses print statements so I'm going to hold off on fixing this for a PR that fixes all of them.

diffusion/models/layers.py
coryMosaicML: Since it uses `mode='nearest'` `align_corners` doesn't have an effect. Antialiasing is probably a good idea, but I left it as default following [the original stable diffusion code](https://github.com/CompVis/stable-diffusion/blob/21f890f9da3cfbeaba8e2ac3c425ee9e998d5229/ldm/modules/diffusionmodules/model.py#L54).

diffusion/models/autoencoder.py
coryMosaicML: Done!

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
coryMosaicML: Done!

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
coryMosaicML: Done!

PR #84, opened 2023-10-16
Update to transforms v2 API

Skylion007:


Skylion007:
Once we update the minimum to PyTorch 2.1, this should be a much simpler PR.

coryMosaicML:
This has lingered for long enough that now it probably makes more sense to update the requirements to PyTorch 2.1 and go with the simpler PR. What do you think?

A-Jacobson:
I believe this is finally handled by #125

PR #94, opened 2023-11-07
Allow masking padding tokens in cross attention layers

jazcollins:
This PR adds a new parameter to the `stable_diffusion_xl` and `stable_diffusion_2` classes called `mask_pad_tokens` that allows for masking out padding tokens in cross attention layers. 

The `generate()` had to get a bit more complicated due to the setting where we pass in pre-tokenized inputs.. we'd now want to allow passing the padding mask with it (as well as for pre-tokenized negative prompts). Let me know if you think of a better way of handling this :/

One small note: this change might be slightly redundant with the `zero_out_negative_prompt` arg (in the generate() function) and `zero_dropped_captions` (in dataloader) that I added not too long ago for zero-ing out empty negative prompts and dropped captions. I think `mask_pad_tokens` ought to serve a similar purpose of masking out the empty text embeddings in the cross attention layers, however `zero_out_negative_prompt/zero_dropped_captions` additionally zero-out the pooled text embedding (used in SDXL as microconditioning). I think we still want to keep that functionality. The cleanest way would prob be to merge these all into one flag?

### DIFF
--- a/diffusion/models/stable_diffusion.py
+++ b/diffusion/models/stable_diffusion.py
@@ -241,7 +241,7 @@ class StableDiffusion(ComposerModel):
 
         # Forward through the model
         return self.unet(noised_latents, timesteps, conditioning,
-                         attention_mask=attention_mask,
+                         encoder_attention_mask=attention_mask,
                          added_cond_kwargs=added_cond_kwargs)['sample'], targets, timesteps
 
     def loss(self, outputs, batch):
--- a/diffusion/models/stable_diffusion.py
+++ b/diffusion/models/stable_diffusion.py
@@ -240,7 +240,9 @@ class StableDiffusion(ComposerModel):
             added_cond_kwargs = {'text_embeds': add_text_embeds, 'time_ids': add_time_ids}
 
         # Forward through the model
-        return self.unet(noised_latents, timesteps, conditioning,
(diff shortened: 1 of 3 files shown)


diffusion/models/stable_diffusion.py
```diff
@@ -422,28 +449,33 @@ def generate(
...
 
-        text_embeddings, pooled_text_embeddings = self._prepare_text_embeddings(prompt, tokenized_prompts,
-                                                                                prompt_embeds, num_images_per_prompt)
+        text_embeddings, pooled_text_embeddings, pad_attn_mask = self._prepare_text_embeddings(
+            prompt, tokenized_prompts, tokenized_prompts_pad_mask, prompt_embeds, num_images_per_prompt)
         batch_size = len(text_embeddings)  # len prompts * num_images_per_prompt
         # classifier free guidance + negative prompts
         # negative prompt is given in place of the unconditional input in classifier free guidance
-        pooled_embeddings = None
+        pooled_embeddings, encoder_attn_mask = pooled_text_embeddings, pad_attn_mask
```
Skylion007, line 457: No reason to do a tuple pack  un pack? Just do two seperate assigns?

coryMosaicML (approved):
LGTM, thanks for this!

PR #95, opened 2023-11-08
Better StreamingDataset defaults while preserving old shuffle settings

snarayan21:
This PR brings the improvements from Streaming v0.7.0 to diffusion while preserving the old shuffle settings that were being used. In detail:

Improvements:
* `relaxed` partitioning is now default. This allows for elastic determinism and resumption on many more numbers of nodes
* `predownload` is now defaulted lower, to `8 * batch_size` to ensure more balanced download demand
* Only when the shuffle algorithm is not `py1s` (the default) or `py2s`, `num_canonical_nodes` will be set to equal the number of physical nodes. Other than that, `num_canonical_nodes` is set the same way as before, to 64 * physical nodes.

Changes:
* `py1s` remains the default shuffle algorithm across this repo. `StreamingDataset` currently defaults to `py1e`. This is so that current workloads don't get messed up, and because `py1e` will be slower because vision datasets are usually much larger than NLP. I can change this after confirmation from you all!
* `shuffle_block_size` remains at 1 << 18 by default. Similar reasoning to above, so that your current workloads don't get messed up.

Happy to discuss these changes with y'all!

Streaming defaults were changed in this commit: https://github.com/mosaicml/streaming/commit/93bf0549bb1d36092a1488004917b594d2ad251e

### DIFF
--- a/diffusion/datasets/coco/coco_captions.py
+++ b/diffusion/datasets/coco/coco_captions.py
@@ -22,23 +22,21 @@ class StreamingCOCOCaption(StreamingDataset):
             Default: ``None``.
         local (str, optional): Local filesystem directory where dataset is cached during operation.
             Default: ``None``.
-        split (str, optional): The dataset split to use. Currently, only ``None`` is supported.
-            Default: ``None``.
         shuffle (bool): Whether to shuffle the samples in this dataset.
             Default: ``False``.
-        tokenizer_name_or_path (str): The name or path of the tokenizer to use.
-            Default: ``'stabilityai/stable-diffusion-2-base'``.
-        transform (Optional[Callable]): The transforms to apply to the image.
-            Default: ``None``.
-        predownload (Optional[int]): The number of samples to prefetch.
-            Default: ``100_000``.
-        download_retry (Optional[int]): The number of times to retry a download.
-            Default: ``2``.
-        download_timeout (Optional[float]): The timeout for a download.
-            Default: ``120``.
+        shuffle_algo (str): What shuffle algorithm to use.
+            Default: ``'py1s'``.
+        shuffle_block_size (int): Unit of shuffling.
+            Default: ``1 << 18``.
         batch_size (Optional[int]):  batch_size that will be used on each device's DataLoader.
             Default: ``None``.
-        image_size (Optional[int]): The size to resize the image to.
+        tokenizer_name_or_path (str): The name or path of the tokenizer to use.
+            Default: ``'stabilityai/stable-diffusion-2-base'``.
+        caption_selection (str): Which caption to use for the image. Must be 'random' or 'first'.
+            Default: ``'first'``.
+        download_timeout (Optional[float]): The timeout for a download.
+            Default: ``120``.
+        transform (Optional[Callable]): The transforms to apply to the image.
             Default: ``None``.
         num_canonical_nodes (int, optional): The number of canonical nodes for shuffle.
             Default: ``None``.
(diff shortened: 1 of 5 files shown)


diffusion/datasets/coco/coco_captions.py
```diff
@@ -50,7 +48,9 @@ def __init__(
         remote,
         local,
         shuffle,
-        batch_size,
+        shuffle_algo: str = 'py1s',
+        shuffle_block_size: int = 1 << 18,
+        batch_size=None,
```
Skylion007, line 53: nt add type hint while we are at it.

diffusion/datasets/image_caption.py
```diff
@@ -65,7 +65,13 @@ def __init__(
         zero_dropped_captions: bool = False,
         **streaming_kwargs,
     ) -> None:
-
+        
+        # Set defaults for vision-friendly streaming args.
+        if 'shuffle_block_size' not in streaming_kwargs:
+            streaming_kwargs['shuffle_block_size'] = 1 << 18
+        if 'shuffle_algo' not in streaming_kwargs:
+            streaming_kwargs['shuffle_algo'] = 'py1s'
```
Skylion007, line 73: Use the setdefault method for python dicts
```suggestion
        streaming_kwargs.setdefault("shuffle_block_size", 1 << 18)
        streaming_kwargs.setdefault("shuffle_algo", "py1s")
```

diffusion/datasets/coco/coco_captions.py
snarayan21: done

diffusion/datasets/image_caption.py
snarayan21: ooh thanks!

snarayan21:
@Skylion007 Sorry this took so long for me to get to! lmk if there are further changes.

diffusion/datasets/coco/coco_captions.py
```diff
@@ -22,23 +22,21 @@ class StreamingCOCOCaption(StreamingDataset):
             Default: ``None``.
         local (str, optional): Local filesystem directory where dataset is cached during operation.
             Default: ``None``.
-        split (str, optional): The dataset split to use. Currently, only ``None`` is supported.
-            Default: ``None``.
         shuffle (bool): Whether to shuffle the samples in this dataset.
             Default: ``False``.
+        shuffle_algo (str): What shuffle algorithm to use.
+            Default: ``'py1s'``.
```
Skylion007, line 28: Out of curiosity, why we do want to use `py1s` is there any reason to?

Skylion007: We are happy to use a different shuffle algo if it's faster.

snarayan21: I wanted to preserve the defaults that worked for you guys in the past, so was keeping `py1s`. `py1e` won't be faster but it'll give you a better shuffle. Since the shuffle is stronger though, it could lead to some slowdowns in some cases, depending on the size of the dataset samples. Because your datasets are so large I think having `py1s` as the default and setting `py1e` if you'd like a stronger shuffle is better.

snarayan21: lmk if any other changes or if you think switching to `py1e` is best.

Skylion007: Oh what would be better but give us a decent shuffle?

snarayan21: If you'd like a better shuffle then setting `py1e` in the yaml should be fine. But jut going to leave this as the default for now.

PR #98, opened 2023-11-20
Hotfix missing lpips requirement

Skylion007:
This hotfixes the missing lpips requirement in setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ install_requires = [
     'triton==2.0.0',
     'torchmetrics[image]==0.11.4',
     'clean-fid',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,6 +17,7 @@
     'triton==2.0.0',
     'torchmetrics[image]==0.11.4',
     'clean-fid',
+    'lpips',
```
coryMosaicML, line 20: Can you pin a version here? Unlikely to change but might save us a headache in the future.

coryMosaicML (approved):

PR #99, opened 2023-11-21
Fixes for doing inference with masked padding

coryMosaicML:
A few inference fixes:

- Pass keyword arguments to inference models
- Add option to disable xformers
- Fix a bug  in constructing masks in the generate function

### DIFF
--- a/diffusion/models/stable_diffusion.py
+++ b/diffusion/models/stable_diffusion.py
@@ -473,12 +473,13 @@ class StableDiffusion(ComposerModel):
             text_embeddings = torch.cat([unconditional_embeddings, text_embeddings])
             if self.sdxl:
                 pooled_embeddings = torch.cat([pooled_unconditional_embeddings, pooled_text_embeddings])  # type: ignore
+            if pad_attn_mask is not None:
+                encoder_attn_mask = torch.cat([uncond_pad_attn_mask, pad_attn_mask])  # type: ignore
         else:
             if self.sdxl:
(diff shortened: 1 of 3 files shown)


diffusion/models/stable_diffusion.py
```diff
@@ -473,12 +473,13 @@ def generate(
...
                 pooled_embeddings = torch.cat([pooled_unconditional_embeddings, pooled_text_embeddings])  # type: ignore
+            if pad_attn_mask is not None:
+                encoder_attn_mask = torch.cat([uncond_pad_attn_mask, pad_attn_mask])  # type: ignore
         else:
             if self.sdxl:
                 pooled_embeddings = pooled_text_embeddings
-
             if pad_attn_mask is not None:
-                encoder_attn_mask = torch.cat([uncond_pad_attn_mask, pad_attn_mask])  # type: ignore
+                encoder_attn_mask = pad_attn_mask
```
jazcollins, line 482: This part shouldn't be necessary since its done on L458 (https://github.com/mosaicml/diffusion/blob/effe5145ec138d14f82f049b1024af015aaa047a/diffusion/models/stable_diffusion.py#L458C9-L458C85)

Sorry about the assigning two variables in one line, it is not the most readable. Prob would be better to break up L458 into:

```
pooled_embeddings = pooled_text_embeddings
encoder_attn_mask = pad_attn_mask
```

coryMosaicML: Should be fixed now!

PR #102, opened 2023-11-28
Make custom autoencoders work with SD2 and SDXL models.

coryMosaicML:
This PR contains a number of QOL changes to make it easier to plug in a custom autoencoder into the existing SD2 and SDXL models. 

Changes:
- AutoEncoder config is now saved as part of the state dict in checkpointing
- Added a `load_autoencoder` function for loading an AutoEncoder object from a composer checkpoint
- Added `LogLatentStatistics` callback for tracking mean and std. dev. of latent channels to ease scaling of the latents
- AutoEncoder now has a `device` function as expected by SD2 and SDXL composer models
- AutoEncoder now returns a latent distribution object as expected by SD2 and SDXL composer models
- SD2 and SDXL models now have the option to download and use custom AutoEncoders by setting `autoencoder_path`
- SD2 and SDXL models now have configurable latent scales and downsampling factors, which may differ with different autoencoders


diffusion/callbacks/log_latent_statistics.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Args:
+        latent_key (str, optional): The key for the latents in the outputs dict. Defaults to 'latents'.
+    """
+
+    def __init__(self, latent_key: str = 'latents'):
+        self.latent_key = latent_key
+        self.counter = 0
+        self.num_channels = 0
+        self.latent_statistics = defaultdict(lambda: 0.0)
```
Skylion007, line 24:
```suggestion
        self.latent_statistics = defaultdict(float)
```

diffusion/models/autoencoder.py
```diff
@@ -307,32 +356,54 @@ def __init__(self,
...
+        config['hidden_channels'] = self.hidden_channels
+        config['latent_channels'] = self.latent_channels
+        config['double_latent_channels'] = self.double_latent_channels
+        config['channel_multipliers'] = self.channel_multipliers
+        config['num_residual_blocks'] = self.num_residual_blocks
+        config['use_conv_shortcut'] = self.use_conv_shortcut
+        config['dropout_probability'] = self.dropout_probability
+        config['resample_with_conv'] = self.resample_with_conv
+        config['use_attention'] = self.use_attention
+        config['zero_init_last'] = self.zero_init_last
```
Skylion007, line 380: Why not just construct this config sttically in one go?

diffusion/callbacks/log_latent_statistics.py
```diff
@@ -0,0 +1,85 @@
...
+        return old_value * self.counter / (self.counter + 1) + new_value / (self.counter + 1)
+
+    def _calc_std(self, mean: float, second_moment: float) -> float:
+        """Compute the standard deviation from the mean and the second moment."""
+        return math.sqrt(second_moment - mean**2)
+
+    def eval_start(self, state: State, logger: Logger):
+        # Reset the counter and wipe the latent statistics
+        self.counter = 0
+        self.latent_statistics = defaultdict(lambda: 0.0)
```
Skylion007, line 46:
```suggestion
        self.latent_statistics = defaultdict(float)
```

diffusion/models/models.py
```diff
@@ -273,7 +351,8 @@ def stable_diffusion_xl(
         model = DeviceGPU().module_to_device(model)
         if is_xformers_installed:
             model.unet.enable_xformers_memory_efficient_attention()
-            model.vae.enable_xformers_memory_efficient_attention()
+            if autoencoder_path is None:
```
Skylion007, line 354: Shouldn't this be a hasattr check instead of autoencoder_path is none check? I don't follow.

diffusion/models/autoencoder.py
```diff
@@ -624,3 +695,47 @@ def forward(self, batch):
...
+    cleaned_state_dict = {}
+    for key in list(state_dict['state']['model'].keys()):
+        if key.split('.')[0] == 'model':
+            cleaned_key = '.'.join(key.split('.')[1:])
+            cleaned_state_dict[cleaned_key] = state_dict['state']['model'][key]
+    # Load the cleaned state dict into the model
+    autoencoder.load_state_dict(cleaned_state_dict, strict=True)
+    # If present, extract the channel means and standard deviations from the state dict
+    if 'LogLatentStatistics' in state_dict['state']['callbacks']:
+        latent_statistics = {'latent_means': [], 'latent_stds': []}
```
Landanjs, line 730: Nit
```suggestion
        latent_statistics = {'latent_channel_means': [], 'latent_channel_stds': []}
```

diffusion/models/models.py
```diff
@@ -33,7 +34,10 @@
 def stable_diffusion_2(
     model_name: str = 'stabilityai/stable-diffusion-2-base',
     pretrained: bool = True,
+    autoencoder_path: Optional[str] = None,
+    autoencoder_local_path: str = '/tmp/autoencoder_weights.pt',
     prediction_type: str = 'epsilon',
+    latent_scale: Optional[float] = None,
```
Landanjs, line 40: Why not?
```suggestion
    latent_scale: Optional[float] = 0.18215,
```

diffusion/models/models.py
```diff
@@ -86,20 +94,50 @@ def stable_diffusion_2(
...
+        # Use the pretrained vae
+        downsample_factor = 8
+        if encode_latents_in_fp16:
+            vae = AutoencoderKL.from_pretrained(model_name, subfolder='vae', torch_dtype=torch.float16)
+        else:
+            vae = AutoencoderKL.from_pretrained(model_name, subfolder='vae')
+        if latent_scale is None:
+            # Default to the SD1/SD2 latent scale
+            latent_scale = 0.18215
+    else:
```
Landanjs, line 115: Do the weights need to specified as `torch_dtype=torch.float16` somewhere for `encode_latents_in_fp16`?

diffusion/models/models.py
Landanjs: Ah, I see.. to have an option to use `latent_statistics` value. But should this be made an explicit argument?  At the expense of another argument

diffusion/models/models.py
Landanjs: +1 hasattr check would be better

Landanjs (commented):
Great work!! Mostly nits and confusion about certain decisions.

2 bigish things:
1. There might be a bug in the `GaussianDistribution` class
2. Is encoding the latents as fp16 missing when loading from the custom AE?

diffusion/callbacks/log_latent_statistics.py
coryMosaicML: Fixed!

diffusion/callbacks/log_latent_statistics.py
coryMosaicML: Fixed!

diffusion/callbacks/log_latent_statistics.py
```diff
@@ -0,0 +1,85 @@
...
+        # Get outputs from the eval forward pass
+        outputs = state.outputs
+        assert isinstance(outputs, dict), 'Outputs must be a dict, got {}'.format(type(outputs))
+        # Get the latents
+        if self.latent_key in outputs:
+            latents = outputs[self.latent_key]  # type: ignore
+        else:
+            raise ValueError(f'Latent key {self.latent_key} not found in outputs.')
+
+        self.num_channels = latents.shape[1]
```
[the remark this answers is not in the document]
coryMosaicML: It turns out that it isn't even necessary to specify this at all, ever. So, it's gone now.

diffusion/models/autoencoder.py
coryMosaicML: Agreed. Fixed it.

diffusion/models/models.py
coryMosaicML: Agreed, it's a `hasattr` now.

diffusion/models/autoencoder.py
```diff
@@ -225,6 +226,52 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    def __getitem__(self, key):
+        if key == 'x_recon':
+            return self.x_recon
+        else:
+            raise KeyError(key)
+
+    @property
+    def sample(self) -> torch.Tensor:
+        """Sample from the output."""
+        return self.x_recon
```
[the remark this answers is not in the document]
coryMosaicML: It is used in our generate function to decode from generated latents [here](https://github.com/mosaicml/diffusion/blob/90b97ed93e42cde98bc536501fd9ef16bb9a9ced/diffusion/models/stable_diffusion.py#L541C17-L541C17). I agree it is a bit weird, and we could modify our `generate` to not use `sample` like this. But this is to make things a bit more compatible with the HF autoencoder. There _could_ be some randomness here in sampling from the VAE decoder, but it would degrade performance.

diffusion/models/autoencoder.py
```diff
@@ -307,32 +356,54 @@ def __init__(self,
...
         nn.init.kaiming_normal_(self.post_quant_conv.weight, nonlinearity='linear')
 
+    @property
+    def device(self) -> torch.device:
+        return next(self.parameters()).device
+
+    def get_extra_state(self):
+        return {'config': self.config}
+
+    def set_extra_state(self, state):
```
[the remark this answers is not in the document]
coryMosaicML: If I understand correctly, PyTorch [expects it](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.set_extra_state).

diffusion/models/autoencoder.py
```diff
@@ -586,7 +657,7 @@ def get_metrics(self, is_train: bool = False):
     def update_metric(self, batch, outputs, metric):
         clamped_imgs = outputs['x_recon'].clamp(-1, 1)
         if isinstance(metric, MeanMetric):
-            metric.update(torch.square(outputs['latents']))
+            metric.update(torch.sqrt(torch.square(outputs['latents'])))
```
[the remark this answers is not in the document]
coryMosaicML: Somewhat of an arbitrary choice, but the quantity with the square root is closer to the value needed for normalizing latents later on. It still isn't quite right though because this gives the mrs instead of the rms. This logging isn't useful if one runs with the `log_latent_statistics` callback, so I could remove it if you think it is misleading.

diffusion/models/autoencoder.py
```diff
@@ -225,6 +226,52 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.var = torch.exp(log_var)
+        self.std = torch.exp(0.5 * log_var)
+
+    def __getitem__(self, key):
+        if key == 'latent_dist':
+            return GaussianDistribution(self.mean, self.log_var)
+        elif key == 'mean':
+            return self.mean[key]
+        elif key == 'log_var':
+            return self.log_var[key]
```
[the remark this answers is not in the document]
coryMosaicML: Yes, it was a bug. Should be fixed now!

diffusion/models/autoencoder.py
```diff
@@ -225,6 +226,52 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+    def __init__(self, mean: torch.Tensor, log_var: torch.Tensor):
+        self.mean = mean
+        self.log_var = log_var
+        self.var = torch.exp(log_var)
+        self.std = torch.exp(0.5 * log_var)
+
+    def __getitem__(self, key):
+        if key == 'latent_dist':
+            return GaussianDistribution(self.mean, self.log_var)
```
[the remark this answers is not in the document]
coryMosaicML: Should return `self`.

diffusion/models/autoencoder.py
coryMosaicML: Fixed!

diffusion/models/models.py
coryMosaicML: Would you be ok with leaving it as `Optional` but having the default be the `0.18215` value? I think that would allow a user to explicitly set `latent_scale = None` to attempt to pick up the `latent_statistics` value.

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: Thanks!

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: Ah, I missed that this was part of a PyTorch `Module`

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
Landanjs: Sorry if I'm missing something, but couldn't this be an `torch.abs` instead of `torch.sqrt(torch.square())`. It doesn't change the code functionally, but the current implementation seems unnecessary to me.

diffusion/models/models.py
Landanjs: I think that would be better.

I still don't like the extra `if latent_scale is None: latent_scale = 0.18215`, but I don't think there is a way to avoid it without adding another argument so it's okay.

diffusion/models/models.py
coryMosaicML: Added an option to load custom autoencoder weights in fp16 for when this option is enabled.

diffusion/models/autoencoder.py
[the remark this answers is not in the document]
coryMosaicML: Ah, I see what you mean now! Yes, it could just be a `torch.abs`. But, given how the mean metric updates stats, I think it is actually better to leave this as a `torch.square` without the `torch.sqrt`.

diffusion/models/models.py
coryMosaicML: Hmm, if you like, the `if latent_scale is None: latent_scale = 0.18215` code could be removed in `model.py` since the `StableDiffusion` class also tries to default to the correct value in the event that `latent_scale` is `None` [here](https://github.com/mosaicml/diffusion/blob/c398359ea28ed0ffd5d7a36723de3b4cb16265fc/diffusion/models/stable_diffusion.py#L117). Though this is just the same ugliness in a different place. 

I'd really like to avoid having to introduce an extra argument for this in any case. What about allowing `latent_scale` to take on a value of `'auto'` to try use the value logged in the autoencoder checkpoint, and leaving the default value the correct number for the base model?

coryMosaicML: I was thinking that in the case of `auto`, we'd only allow `latent_scale` to be a `float` (defaulting to the correct value) or a `str`, and throw an error if it's `auto` without an autoencoder checkpoint to load stats from. So, the check for `None` could be removed. I think a setting for `latent_statistics` would work similarly. Definitely open to making it more specific. I'd rather get this right the first time than having something else to clean up later.

Landanjs: To me, `'latent_statistics'` is a lot clearer than `'auto'`

coryMosaicML: Ok, I made this change!

diffusion/models/autoencoder.py
```diff
@@ -307,32 +356,54 @@ def __init__(self,
...
-        return {'mean': mean, 'log_var': log_var}
+        return GaussianDistribution(mean, log_var)
 
-    def decode(self, z: torch.Tensor) -> Dict[str, torch.Tensor]:
+    def decode(self, z: torch.Tensor) -> AutoEncoderOutput:
         """Decode a latent tensor into an output tensor."""
         z = self.post_quant_conv(z)
         x_recon = self.decoder(z)
-        return {'x_recon': x_recon}
+        return AutoEncoderOutput(x_recon)
```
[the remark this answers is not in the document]
coryMosaicML: Yes, they do. Do you think it's needed here? I was aiming for the minimal implementation that's compatible with our current model classes rather than compatibility with HF. A cleaner option could be to change our model classes to use the HF `return_dict` and then switch this back to dict outputs.

A-Jacobson (approved):
Seems that you've addressed aaron and landans requests already. Approving and will cut a new version once this is merged.

PR #104, opened 2023-12-08
LogDiffusionImages Features and Refactors

Landanjs:
Adjustments to `LogDiffusionImages`

Features:
1. `batch_size` argument: batches the `prompts`, so we can generate more images at higher resolution
2. `size` argument accepts a tuple: allows us to specify non-square resolutions. I could make this a list of tuples if desired and we are okay with the complexity

Refactors (because I couldn't stop myself)
1. Removed `text_key` argument: we only used this to get the caption device and use this to move `tokenized_prompts`, but moving `tokenized_prompts` to the device should be handled in `generate()`
2. Changed the callback to run at `eval_start` instead of `eval_batch_end`. `eval_batch_end` required an if-statement to prevent logging images multiple times.

### DIFF
--- a/diffusion/callbacks/log_diffusion_images.py
+++ b/diffusion/callbacks/log_diffusion_images.py
@@ -3,7 +3,7 @@
 
 """Logger for generated images."""
 
-from typing import List, Optional
+from typing import List, Optional, Tuple, Union
 
 import torch
 from composer import Callback, Logger, State
@@ -19,7 +19,8 @@ class LogDiffusionImages(Callback):
 
     Args:
         prompts (List[str]): List of prompts to use for evaluation.
-        size (int, optional): Image size to use during generation. Default: ``256``.
+        size (int, Tuple[int, int], optional): Image size to use during generation.
+            If using a tuple, specify as (height, width).  Default: ``256``.
         num_inference_steps (int, optional): Number of inference steps to use during generation. Default: ``50``.
         guidance_scale (float, optional): guidance_scale is defined as w of equation 2
             of the Imagen Paper. Guidance scale is enabled by setting guidance_scale > 1.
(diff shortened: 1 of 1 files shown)


diffusion/callbacks/log_diffusion_images.py
```diff
@@ -19,7 +19,10 @@ class LogDiffusionImages(Callback):
 
     Args:
         prompts (List[str]): List of prompts to use for evaluation.
-        size (int, optional): Image size to use during generation. Default: ``256``.
+        size (int, Tuple[int, int], optional): Image size to use during generation.
+            If using a tuple, specify as (height, width).  Default: ``256``.
+        batch_size (int, optional): The batch size of the prompts passed to the generate function. If not specified,
```
coryMosaicML, line 24: Would it be better to set the default batch size to be 1 if this isn't specified so an OOM due to misconfiguring is less likely?

diffusion/callbacks/log_diffusion_images.py
```diff
@@ -38,62 +40,67 @@ class LogDiffusionImages(Callback):
 
     def __init__(self,
                  prompts: List[str],
-                 size: Optional[int] = 256,
+                 size: Optional[Union[Tuple[int, int], int]] = 256,
```
coryMosaicML, line 43: This shouldn't have the `Optional` I think. A few others like this too here, if you want to fix them.

diffusion/callbacks/log_diffusion_images.py
```diff
@@ -38,62 +40,67 @@ class LogDiffusionImages(Callback):
...
-                model = state.model.module
+    def eval_start(self, state: State, logger: Logger):
+        # Get the model object if it has been wrapped by DDP
+        # We need this to access the text keys, tokenizer, and image generation function.
+        if isinstance(state.model, DistributedDataParallel):
+            model = state.model.module
+        else:
+            model = state.model
+
+        if self.tokenized_prompts is None:
```
coryMosaicML, line 69: I have always found this a bit ugly. What do you think about not tokenizing the prompts here, and instead letting `model.generate` handle that by passing in string prompts? It would allow us to avoid the check for `model.sdxl` here, and avoid making the assumption that we can call `model.tokenizer` this way. Would require slightly different batching below though.

coryMosaicML (approved):
Looks good! Made a couple suggestions.

diffusion/callbacks/log_diffusion_images.py
Landanjs: I removed optional from size and guidance_scale, then added a type to num_inference_steps. Let me know if others should be removed!

diffusion/callbacks/log_diffusion_images.py
Landanjs: Sounds good, it shouldn't slow eval too much and is better than the run crashing

diffusion/callbacks/log_diffusion_images.py
Landanjs: Removed! It's a lot cleaner

PR #108, opened 2024-01-03
Add option to use E5 text encoder for SDXL

A-Jacobson:
adds a simple switch to use e5 text encoder with SDXL. this is accomplished by splicing e5 into our joint text encoder class. Currently, this approach has some limitations:
- I wanted to avoid changing the API + building out a large registry or text encoders so only e5-large-v2 is currently supported. (t5 and other text encoders i tester were not supported by automodel/autotokenizer so I opted to keep it simple for now and leave them out.
- decided to stick with openclip vs openai clip as the HF model supports the projection layer needed for SDXL out of the box
- truncate sequence max-length to 77 (clip max len) vs 512 (e5 max length).

~to enable e5 add `use_e5: true` to both your dataset and model.~

after feedback, `model_name` and `tokenizer_name_or_path` now need to be updated to `sdxl-e5` to enable e5 training.

### DIFF
--- a/diffusion/datasets/image_caption.py
+++ b/diffusion/datasets/image_caption.py
@@ -43,7 +43,7 @@ class StreamingImageCaptionDataset(StreamingDataset):
         image_key (str): Key associated with the image in the streaming dataset. Default: ``'image'``.
         caption_key (str): Key associated with the caption in the streaming dataset. Default: ``'caption'``.
         sdxl (bool): Whether or not we're training SDXL. Default: `False`.
-        use_e5 (bool): Whether to use e5-large-v2 tokenizer instead of openai CLIP. 
+        use_e5 (bool): Whether to use e5-large-v2 tokenizer instead of openai CLIP.
             Enable if using e5 text encoder. Defaults to False.
         zero_dropped_captions (bool): If True, zero out text embeddings for dropped captions. Default: ``False``.
 
@@ -202,6 +202,8 @@ def build_streaming_image_caption_dataloader(
         local (str, Sequence[str]): One or more local filesystem directories where dataset is cached during operation.
         batch_size (int): The batch size to use for both the ``StreamingDataset`` and ``DataLoader``.
         tokenizer_name_or_path (str): The name or path of the tokenizer to use. Default: ``'stabilityai/stable-diffusion-2-base'``.
+        use_e5 (bool): Whether to use e5-large-v2 tokenizer instead of openai CLIP.
+            Enable if using e5 text encoder. Defaults to False.
         caption_drop_prob (float): The probability of dropping a caption. Default: ``0.0``.
         microcond_drop_prob (float): The probability of dropping microconditioning. Only relevant for SDXL. Default: ``0.0``.
         resize_size (int): The size to resize the image to. Default: ``256``.
--- a/diffusion/datasets/image_caption.py
+++ b/diffusion/datasets/image_caption.py
@@ -32,7 +32,9 @@ class StreamingImageCaptionDataset(StreamingDataset):
             ``StreamingImageCaptionDataset`` uses either ``streams`` or ``remote``/``local``. Default:``None``.
         remote (str, optional): Remote directory (S3 or local filesystem) where dataset is stored. Default: ``None``.
         local (str, optional): Local filesystem directory where dataset is cached during operation. Default: ``None``.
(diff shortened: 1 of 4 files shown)


A-Jacobson:
@jazcollins if you could confirm i didn't horrendously break the tokenizers or sdxl code you wrote that would be great =)

diffusion/datasets/image_caption.py
```diff
@@ -43,6 +43,8 @@ class StreamingImageCaptionDataset(StreamingDataset):
         image_key (str): Key associated with the image in the streaming dataset. Default: ``'image'``.
         caption_key (str): Key associated with the caption in the streaming dataset. Default: ``'caption'``.
         sdxl (bool): Whether or not we're training SDXL. Default: `False`.
+        use_e5 (bool): Whether to use e5-large-v2 tokenizer instead of openai CLIP.
```
Skylion007, line 46: Having this is a boolean has bad code smell, is there anyway we can automatically enable  this boolean based on the sub_tokenizer? or by adding a substring to the end of the tokenizer maybe?

diffusion/models/models.py
```diff
@@ -219,8 +222,10 @@ def stable_diffusion_xl(
         unet = UNet2DConditionModel.from_pretrained(unet_model_name, subfolder='unet')
     else:
         config = PretrainedConfig.get_config_dict(unet_model_name, subfolder='unet')
+        if use_e5:
+            # e5 + clip embedding dims + micro conditioning
+            config[0]['cross_attention_dim'] = 2304
```
Skylion007, line 227: Should we just make this configurable as a seperate arg instead of having it overridden with a boolean?

jazcollins (commented):
Overall looks good to me aside from some small suggestions to remove the `use_e5` flag!

Also - do we want to truncate e5 tokenizer by the CLIP tokenizer's max_length? As the code is currently is currently written - yes, we have to, because we stack the tokenized outputs in the dataloader. However, we don't have to do that, and could potentially have different length tokenized outputs for the two text encoders if that makes sense to do.

diffusion/models/models.py
```diff
@@ -238,8 +243,8 @@ def stable_diffusion_xl(
     except:  # for handling SDXL vae fp16 fixed checkpoint
         vae = AutoencoderKL.from_pretrained(vae_model_name, torch_dtype=torch_dtype)
 
-    tokenizer = SDXLTokenizer(model_name)
-    text_encoder = SDXLTextEncoder(model_name, encode_latents_in_fp16)
+    tokenizer = SDXLTokenizer(model_name, use_e5=use_e5)
+    text_encoder = SDXLTextEncoder(model_name, encode_latents_in_fp16, use_e5=use_e5)
```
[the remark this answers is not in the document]
A-Jacobson: definitely an option. I wasnt sure about this because sdxl-e5 isnt a huggingface model path. cleaner interface though!

A-Jacobson:
> Also - do we want to truncate e5 tokenizer by the CLIP tokenizer's max_length? As the code is currently is currently written - yes, we have to, because we stack the tokenized outputs in the dataloader. However, we don't have to do that, and could potentially have different length tokenized outputs for the two text encoders if that makes sense to do.

I believe they have to be the same length because they're concatenated on the embedding dim not the sequence dim later on. We can't do concatenation unless we either pad clip to 512 or truncate e5 to 77. e5 uses wordpiece and clip uses BPE so I THINK the # of tokens per prompt should be similar.

diffusion/models/models.py
A-Jacobson: I'm going to leave this as is to avoid the mental burden of a user having know or figure out what the xattention dim should be if they enable e5. It's ugly, and something that would ideally be enabled via it's own unique HF checkpoint but mocking said behavior like this is probably the best we can do atm.

coryMosaicML (approved):

diffusion/models/models.py
```diff
@@ -311,6 +313,8 @@ def stable_diffusion_xl(
             # Adapt the unet config to account for differing number of latent channels if necessary
             unet_config['in_channels'] = vae.config['latent_channels']
             unet_config['out_channels'] = vae.config['latent_channels']
+        if model_name == 'sdxl-e5':  # e5 + clip embedding dims + micro conditioning
+            unet_config]['cross_attention_dim'] = 2304
```
A-Jacobson, line 317:
```suggestion
            unet_config['cross_attention_dim'] = 2304
```

fix missing brackets

A-Jacobson:
handled by #124

PR #111, opened 2024-01-09
Remove rounding in aspect ratio bucketing transform

Landanjs:
I initially used `torch.round` since I used rounded aspect ratio buckets for the first LAION bucketing. With the updated LAION and Shutterstocks, there are no more rounded buckets so `torch.round` needs to be removed.


jazcollins (approved):
lgtm thanks!

PR #112, opened 2024-01-10
Add sample SDXL yamls and update README

jazcollins:
Despite our repo supporting SDXL, I realized we have no documentation on how to use it! This PR provides sample SDXL yamls for each resolution of training and updates the main README with info on how to train SDXL.


yamls/mosaic-yamls/SDXL-base-1024.yaml
```diff
@@ -0,0 +1,125 @@
...
+    memory_monitor:
+      _target_: composer.callbacks.memory_monitor.MemoryMonitor
+    runtime_estimator:
+      _target_: composer.callbacks.runtime_estimator.RuntimeEstimator
+    optimizer_monitor:
+      _target_: composer.callbacks.OptimizerMonitor
+  trainer:
+    _target_: composer.Trainer
+    device: gpu
+    max_duration: 200000ba
```
A-Jacobson, line 115: 200k iters at 1024 is a lot. is this the most recent schedule we want to recommend? the sdxl paper didn't have details on this stage if i remember correctly.

A-Jacobson (approved):
Nice! long overdue.

Added one comment about the 1024 schedule, if that's correct then feel free to merge.

yamls/mosaic-yamls/SDXL-base-1024.yaml
jazcollins: Correct, the paper doesn't mention this detail. We might very likely be able to get away with less, but 200K is what we have been doing!

coryMosaicML (approved):
Looks good! Thank you for doing this.

PR #115, opened 2024-01-20
Add algorithm to control randomness over different eval times

coryMosaicML:
As is, during eval we get different randomness every time we run eval. This leads to eval loss curves having extra noise. This PR adds an algorithm to pin the eval randomness such that we get the same randomness per sample for every eval, provided number of devices doesn't change.

A comparison of an eval loss curve with/without controlling the randomness over a few resumptions:
Screenshot 2024-01-19 at 4 07 16 PM

To use it, add the algorithm to the `algorithms` section of the yaml
Screenshot 2024-01-19 at 4 09 23 PM

### DIFF
--- a/diffusion/algorithms/pin_diffusion_rng.py
+++ b/diffusion/algorithms/pin_diffusion_rng.py
@@ -3,7 +3,7 @@
 
 """Algorithm to pin diffusion process noise."""
 
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import torch
 from composer.core import Algorithm, Event, State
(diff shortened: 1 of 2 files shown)


diffusion/algorithms/pin_diffusion_rng.py
```diff
@@ -0,0 +1,70 @@
...
+
+from diffusion.models import StableDiffusion
+
+
+class PinDiffusionRNG(Algorithm):
+    """Algorithm to pin diffusion process noise."""
+
+    def __init__(self) -> None:
+        self.seed = 0
+        self.train_rng_state = None
```
Skylion007, line 21: Should be typed, implicit optionals are bad.

A-Jacobson (approved):
Approving. I think it's ok to address aaron's typing detail and merge at your leasure. This is a low-danger PR with an optional algorithm that's sat around too long.

diffusion/algorithms/pin_diffusion_rng.py
coryMosaicML: Fixed!

PR #117, opened 2024-01-30
Add option to finetune models with LoRA

coryMosaicML:
Adds an option to do LoRA finetuning for SD2 and SDXL models.

One can use it by passing the desired `lora_rank` in the config, like so:
```
model:
  _target_: diffusion.models.models.stable_diffusion_xl
  clip_qkv: null
  pretrained: true
  lora_rank: 64
  precomputed_latents: false
  encode_latents_in_fp16: true
  fsdp: true
  val_metrics:
  - _target_: torchmetrics.MeanSquaredError
  val_guidance_scales: []
  loss_bins: []
```
By default this is off, and LoRA will only be used if a `lora_rank` is set.

To use it with a custom checkpoint, pass the desired `load_path` to the trainer, and also set
```
load_weights_only: true
load_strict_model_weights: false
```

### DIFF
--- a/diffusion/models/models.py
+++ b/diffusion/models/models.py
@@ -10,7 +10,6 @@ from typing import List, Optional, Tuple, Union
 import torch
 from composer.devices import DeviceGPU
 from diffusers import AutoencoderKL, DDIMScheduler, DDPMScheduler, EulerDiscreteScheduler, UNet2DConditionModel
-from diffusers.loaders import AttnProcsLayers
 from diffusers.models.attention_processor import LoRAAttnProcessor
 from torchmetrics import MeanSquaredError
 from torchmetrics.image.fid import FrechetInceptionDistance
@@ -195,11 +194,8 @@ def stable_diffusion_2(
         model.unet.set_attn_processor(attn_processor)
 
     if lora_rank is not None:
-        lora_layers = make_lora_layers(unet, lora_rank)
-        trainable_params = lora_layers.parameters()
-    else:
-        trainable_params = model.parameters()
-    return model, trainable_params
+        make_lora_layers(unet, lora_rank)
+    return model
 
 
 def stable_diffusion_xl(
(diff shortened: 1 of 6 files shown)


diffusion/models/models.py
```diff
@@ -605,3 +621,29 @@ def __call__(self, prompt, padding, truncation, return_tensors, max_length=None)
         for key in tokenized_output.keys():
             tokenized_output[key] = [tokenized_output[key], tokenized_output_2[key]]
         return tokenized_output
+
+
+def make_lora_layers(unet, rank):
+    """From https://huggingface.co/docs/diffusers/training/lora."""
+    lora_attn_procs = {}
+    for name in unet.attn_processors.keys():
```
Landanjs, line 629: Do the `attn_processors` include the self attention layers or are those contained in a separate attribute? Curious if this is only adjusting cross attention layers (as intended).

Landanjs (commented):
Thanks for adding this!!

Instead of having every model return `trainable_params` as a second argument, could we check for a LoRA config argument in `train.py`, then extract the relevant trainable parameters? For example:

```
if getattr(config.model, 'lora_rank'):
    optimizer = hydra.utils.instantiate(config.optimizer, params=model.attn_processors.parameters())
```

To me, requiring the model factory functions to return a tuple is very error prone (e.g. the number of lines you had to change to accommodate this requirement)

diffusion/models/models.py
coryMosaicML: I'm not sure, but this is the way huggingface suggests LoRA be done (see the link) so I think it is safe to do it this way.

Landanjs:
Hmmm... This would be fragile if we start using `unet` models without `attn_processors` attribute (is this likely?). What do you think about setting the LoRA parameters to be a part of the ComposerModel in addition to setting it in the unet? Then in `train.py`, you can specifically check for `model.lora_layers`

```
self.lora_layers = make_lora_layers(unet, lora_rank)
```

coryMosaicML:
Closing this because diffusers API has changed.

PR #118, opened 2024-02-09
Add image only dataset + script to add captions generated by LLaVA to a streaming dataset

coryMosaicML:
The image only dataset exists to allow captioning datasets that do not already have captions in them, and to provide the option to return all fields as part of sampling. 

The captioning script can be run on a multi-gpu instance like so:
```
composer scripts/batched-llava-caption.py --remote REMOTE_FOR_ORIGINAL_DATASET --local LOCAL_SHARDS_DIR --output REMOTE_FOR_NEW_DATASET
```

### DIFF
--- a/diffusion/datasets/image.py
+++ b/diffusion/datasets/image.py
@@ -1,0 +1,134 @@
+# Copyright 2022 MosaicML Diffusion authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Streaming Image dataset."""
+
+import logging
+from io import BytesIO
+from typing import Callable, Dict, List, Optional, Sequence, Union
+
+from PIL import Image
+from streaming import Stream, StreamingDataset
+from torch.utils.data import DataLoader
+from torchvision import transforms
+
+log = logging.getLogger(__name__)
+
+# Disable PIL max image size limit
+Image.MAX_IMAGE_PIXELS = None
+
+
+class StreamingImageDataset(StreamingDataset):
+    """Streaming dataset for images.
+
+    Args:
+        streams (Sequence[Stream], optional): One or more Streams to stream/cache samples from.
+            ``StreamingImageCaptionDataset`` uses either ``streams`` or ``remote``/``local``. Default:``None``.
+        remote (str, optional): Remote directory (S3 or local filesystem) where dataset is stored. Default: ``None``.
+        local (str, optional): Local filesystem directory where dataset is cached during operation. Default: ``None``.
+        transform (Callable, optional): The transforms to apply to the image. Default: ``None``.
+        image_key (str): Key associated with the image in the streaming dataset. Default: ``'image'``.
+
+        **streaming_kwargs: Additional arguments to pass in the construction of the StreamingDataloader
+    """
+
+    def __init__(
+        self,
+        streams: Optional[Sequence[Stream]] = None,
+        remote: Optional[str] = None,
+        local: Optional[str] = None,
+        transform: Optional[Callable] = None,
+        image_key: str = 'image',
+        **streaming_kwargs,
+    ) -> None:
+
+        # Set defaults for vision-friendly streaming args.
+        streaming_kwargs.setdefault('shuffle_block_size', 1 << 18)
+        streaming_kwargs.setdefault('shuffle_algo', 'py1s')
+
+        super().__init__(
+            streams=streams,
+            remote=remote,
+            local=local,
+            **streaming_kwargs,
+        )
+
+        self.transform = transform
+        self.image_key = image_key
+
+    def __getitem__(self, index):
+        sample = super().__getitem__(index)
+        # Image
+        img = sample[self.image_key]
+        if not isinstance(img, Image.Image):
+            img = Image.open(BytesIO(sample[self.image_key]))
+        if img.mode != 'RGB':
+            img = img.convert('RGB')
+        # Image transforms
+        if self.transform is not None:
+            img = self.transform(img)
+        return {self.image_key: img}
+
+
+def build_streaming_image_dataloader(
+    remote: Union[str, List],
+    local: Union[str, List],
+    batch_size: int,
+    transform: Optional[List[Callable]] = None,
+    image_key: str = 'image',
+    streaming_kwargs: Optional[Dict] = None,
+    dataloader_kwargs: Optional[Dict] = None,
+):
+    """Builds a streaming LAION dataloader.
+
+    Args:
+        remote (str, Sequence[str]): One or more remote directories (S3 or local filesystem) where dataset is stored.
+        local (str, Sequence[str]): One or more local filesystem directories where dataset is cached during operation.
+        batch_size (int): The batch size to use for both the ``StreamingDataset`` and ``DataLoader``.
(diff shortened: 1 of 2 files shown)


diffusion/datasets/image.py
```diff
@@ -0,0 +1,150 @@
...
+        self.transform = transform
+        self.image_key = image_key
+        self.image_output_key = image_output_key if image_output_key is not None else image_key
+        self.return_all_fields = return_all_fields
+
+    def __getitem__(self, index):
+        sample = super().__getitem__(index)
+        # Image
+        if not isinstance(sample[self.image_key], Image.Image):
+            img = Image.open(BytesIO(sample[self.image_key])).copy()
```
Skylion007, line 71: Why the copy here?

diffusion/datasets/image.py
```diff
@@ -0,0 +1,150 @@
...
+        streaming_kwargs = {}
+    if dataloader_kwargs is None:
+        dataloader_kwargs = {}
+
+    # Check types for remote and local
+    if isinstance(remote, str) and isinstance(local, str):
+        remote, local = [remote], [local]
+    elif isinstance(remote, Sequence) and isinstance(local, Sequence):
+        if len(remote) != len(local):
+            ValueError(
```
Skylion007, line 120: This is suppose to raise ValueErrors here, not just instantiate them, right?

coryMosaicML: This is to allow passing in a single `remote` + `local` pair as a string instead of a list of multiple, so it shouldn't throw an error. Though questionable if this is how we'd like this implemented.

Skylion007: I meant:
```suggestion
            raise ValueError(
```

coryMosaicML: Oops! Yes you're right. Nice catch!

diffusion/datasets/image.py
coryMosaicML: It's uneccesary. Removed it now!

scripts/batched-llava-caption.py
```diff
@@ -0,0 +1,344 @@
...
+    from llava.constants import DEFAULT_IMAGE_TOKEN  # type: ignore
+    from llava.constants import DEFAULT_IM_END_TOKEN, DEFAULT_IM_START_TOKEN, IMAGE_TOKEN_INDEX  # type: ignore
+    from llava.conversation import SeparatorStyle, conv_templates  # type: ignore
+    from llava.mm_utils import get_model_name_from_path, tokenizer_image_token  # type: ignore
+    from llava.model.builder import load_pretrained_model  # type: ignore
+    from llava.utils import disable_torch_init  # type: ignore
+except ImportError:
+    raise ImportError(
+        'LLaVA is not installed. Please install it with `pip install llava@git+https://github.com/haotian-liu/LLaVA.git`'
+    )
```
Skylion007, line 29: Use `raise from` statement here to explicitly chain the exceptions

scripts/batched-llava-caption.py
```diff
@@ -0,0 +1,344 @@
...
+                 device: Optional[torch.device] = None):
+        self.model_name = model_name
+        self.conv_mode = 'llava_v1'
+        self.max_tokens = max_tokens
+        self.device = torch.device('cuda') if device is None else device
+
+        self.tokenizer, self.model, self.image_processor, self.context_len = self.load_llava(quantize=quantize,
+                                                                                             multi_gpu=multi_gpu)
+        self.generate = self.model.generate
+        if compile:
```
Skylion007, line 56: This api doesn't allow us to pass custom dict with custom compile options

scripts/batched-llava-caption.py
```diff
@@ -0,0 +1,344 @@
...
+            self.model = torch.compile(self.model)
+            self.generate = torch.compile(self.generate)
+
+        self.input_ids: Optional[torch.Tensor] = None
+
+    def to(self, device: torch.device):
```
Skylion007, line 62: Why not just make this a `nn.Module` then you don't have to define `.to`

Landanjs (approved):
From a quick skim, LGTM! I didn't go through the logic in detail, but I can do it if needed. I did not look at the llava model since I'm not familiar, would be better for @A-Jacobson or @jazcollins to check that specific part.

diffusion/datasets/image.py
```diff
@@ -0,0 +1,151 @@
...
+        self.image_output_key = image_output_key if image_output_key is not None else image_key
+        self.return_all_fields = return_all_fields
+
+    def __getitem__(self, index):
+        sample = super().__getitem__(index)
+        # Image
+        if not isinstance(sample[self.image_key], Image.Image):
+            img = Image.open(BytesIO(sample[self.image_key]))
+        else:
+            img = sample[self.image_key].copy()
```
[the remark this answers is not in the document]
coryMosaicML: Copy here was to prevent modification of the data in `sample[image_key]` by subsequent transforms since that gets returned if `return_all_fields=True`. Not sure if there is ever a case where this is an issue, so I can remove it if you think it's unnecessary.

scripts/batched-llava-caption.py
coryMosaicML: True, was trying to balance simplicity with flexibility here, but open to a suggestion to allow a custom config here.

diffusion/datasets/image.py
```diff
@@ -0,0 +1,151 @@
...
+        if img.mode != 'RGB':
+            img = img.convert('RGB')
+        # Image transforms
+        if self.transform is not None:
+            img = self.transform(img)
+        # Make the output
+        if self.image_output_key != self.image_key and self.image_output_key in sample:
+            raise ValueError(f'Output key {self.image_output_key} already exists in the sample.')
```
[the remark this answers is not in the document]
coryMosaicML: My thinking here was that one could accidentally overwrite data by specifying an output key the same as a key in the sample which could be really undesirable for something downstream and potentially difficult to debug, so I prefer to just break here rather than raise a warning that could be missed. You're right that this probably only needs to happen if the key in question is being returned though.

scripts/batched-llava-caption.py
coryMosaicML: Done!

scripts/batched-llava-caption.py
coryMosaicML: `.to` here isn't used, so I just removed it.

scripts/batched-llava-caption.py
```diff
@@ -0,0 +1,344 @@
...
+    device = torch.device(f'cuda:{dist.get_local_rank()}' if torch.cuda.is_available() else 'cpu')
+    captioner = LLaVACaptioner(model_name=args.model_name,
+                               max_tokens=args.max_tokens,
+                               compile=args.compile,
+                               quantize=args.quantize,
+                               multi_gpu=args.multi_gpu,
+                               device=device)
+
+    # Each rank needs it's own output
+    output_dir = args.output + f'/{dist.get_global_rank()}'
```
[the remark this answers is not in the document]
coryMosaicML: Fixed!

PR #125, opened 2024-03-05
Update to latest transformers, diffusers, and other packages.

coryMosaicML:
This PR updates versions in `setup.py` to latest package versions where appropriate. Code changes are mostly due to changes in the diffusers API and return types.


Skylion007 (approved):

diffusion/models/layers.py
```diff
@@ -14,8 +14,10 @@
 except:
     pass
 
+T = TypeVar('T', bound=nn.Module)
```
Skylion007, line 17: This is a private a typevar, should be prefixed with a underscore.

coryMosaicML: Good point! Fixed now!

PR #126, opened 2024-03-12
arbitrary aspect ratio buckets

Landanjs:


diffusion/datasets/laion/transforms.py
```diff
@@ -47,17 +48,9 @@ def __call__(self, img):
 class RandomCropAspectRatioTransorm:
     """Assigns an image to a pre-defined set of aspect ratio buckets, then resizes and crops to fit into the bucket."""
 
-    def __init__(self):
-        self.height_buckets = torch.tensor([
```
coryMosaicML, line 51: I think we should either default or document these height/width buckets somewhere. Otherwise it will be a pain to specify if we ever need to compare to some of our old runs. Not urgent though so this does not need to hold up merging.

coryMosaicML (approved):
Looks good! Thanks for making this customizable!

PR #128, opened 2024-03-13
make local paths optional

A-Jacobson:
This PR is aimed at cleaning up our yamls.  It makes local paths optional and passes tokenizers directly to datasets in `train.py` so the tokenizer string doesn't have to be specified in multiple places.

The default local path with be `/tmp/<streaming path>` the streaming path with have the cloud provide prefix removed (oci://, s3://, ect).

For models that don't have tokenizer attributes (autoencoders), no tokenizer will be used and raw text will be returned in the batch.

It doesn't currently support the old laion streaming dataset or coco_captions dataset.

tested with small sdxl-256 runs.

### DIFF
--- a/diffusion/train.py
+++ b/diffusion/train.py
@@ -73,6 +73,7 @@ def train(config: DictConfig) -> None:
     # upstream composer changes that could make this easier.
     train_dataloader: Union[Iterable, DataSpec, Dict[str, Any]] = hydra.utils.instantiate(
         config.dataset.train_dataset,
+        tokenizer=model.tokenizer,
         batch_size=config.dataset.train_batch_size // dist.get_world_size(),
     )
     # Need to sleep for a bit to avoid dataloader crash
@@ -88,7 +89,8 @@ def train(config: DictConfig) -> None:
             print(OmegaConf.to_yaml(eval_conf))
             eval_dataloader = hydra.utils.instantiate(
                 eval_conf.eval_dataset,
-                config.dataset.eval_batch_size // dist.get_world_size(),
+                tokenizer=model.tokenizer,
+                batch_size=config.dataset.eval_batch_size // dist.get_world_size(),
             )
             evaluator = hydra.utils.instantiate(eval_conf.evaluator, dataloader=eval_dataloader)
             # Need to sleep for a bit to avoid dataloader crash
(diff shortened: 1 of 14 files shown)


diffusion/train.py
```diff
@@ -65,14 +65,17 @@ def train(config: DictConfig) -> None:
...
     else:
         optimizer = hydra.utils.instantiate(config.optimizer, params=model.parameters())
+        tokenizer = model.tokenizer
 
     # Load train dataset. Currently this expects to load according to the datasetHparam method.
     # This means adding external datasets is currently not super easy. Will refactor or check for
     # upstream composer changes that could make this easier.
     train_dataloader: Union[Iterable, DataSpec, Dict[str, Any]] = hydra.utils.instantiate(
         config.dataset.train_dataset,
+        tokenizer=tokenizer,
```
coryMosaicML, line 78: I think it might be a good idea to only pass tokenizer to the dataloader if `model.tokenizer` exists. Reason being we also have an image only dataset [here](https://github.com/mosaicml/diffusion/blob/da716d7c424d58f1d8e6f2e91c6f0b4029ea53ae/diffusion/datasets/image.py#L21) that doesn't take a tokenizer, which is what one _should_ use for autoencoder training but the yamls are out of data (my fault, sorry...)

Any other solution that doesn't require dataloaders to accept a tokenizer argument would also work.

diffusion/train.py
```diff
@@ -88,7 +91,8 @@ def train(config: DictConfig) -> None:
             print(OmegaConf.to_yaml(eval_conf))
             eval_dataloader = hydra.utils.instantiate(
                 eval_conf.eval_dataset,
-                config.dataset.eval_batch_size // dist.get_world_size(),
+                tokenizer=tokenizer,
```
coryMosaicML, line 94: Same as above

diffusion/train.py
```diff
@@ -99,6 +103,7 @@ def train(config: DictConfig) -> None:
 
     else:
         eval_set = hydra.utils.instantiate(config.dataset.eval_dataset,
+                                           tokenizer=model.tokenizer,
```
coryMosaicML, line 106: Same as above

coryMosaicML (approved):
Looks great! Thank you for cleaning this up, this is much nicer. Left one minor suggestion.

diffusion/train.py
A-Jacobson: Added 3 ugly If/else branches =(. 

I tried to get around this by adding the tokenizer directly to the omegaconf config, but omegaconf doesn't like having random giant objects added to it.

the functionality should be there now though. 

Also updated eval.py

PR #129, opened 2024-03-14
Add option to specify image output key in image dataloader factory

coryMosaicML:


A-Jacobson (approved):
Oh hey, there's the hidden Image only dataset. I wonder who added that!

PR #132, opened 2024-03-28
Add option to set per-channel mean, std. dev. of the autoencoder latents when training the UNet

coryMosaicML:
This PR adds the option of setting per channel values for the normalization of the autoencoder latents. By default, the scaling uses zero mean and the same scaling value for all channels, which is the previous behavior. One can override this with a scalar for one or both of `latent_mean` or `latent_std` which will use that value for all channels, or specify a tuple of per-channel values. One can also set `latent_mean='latent_statistics'` and/or `latent_std='latent_statistics'` to use the value from the autoencoder checkpoint if it has tracked latent statistics.

### DIFF
--- a/diffusion/models/models.py
+++ b/diffusion/models/models.py
@@ -138,9 +138,9 @@ def stable_diffusion_2(
             unet_config['out_channels'] = vae.config['latent_channels']
         # Init the unet from the config
         unet = UNet2DConditionModel(**unet_config)
-    if type(latent_mean) is float:
+    if isinstance(latent_mean, float):
         latent_mean = (latent_mean,) * unet_config['in_channels']
-    if type(latent_std) is float:
+    if isinstance(latent_std, float):
         latent_std = (latent_std,) * unet_config['in_channels']
     assert isinstance(latent_mean, tuple) and isinstance(latent_std, tuple)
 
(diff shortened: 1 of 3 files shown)


diffusion/models/models.py
```diff
@@ -187,7 +205,8 @@ def stable_diffusion_xl(
     autoencoder_path: Optional[str] = None,
     autoencoder_local_path: str = '/tmp/autoencoder_weights.pt',
     prediction_type: str = 'epsilon',
-    latent_scale: Union[float, str] = 0.13025,
+    latent_mean: Union[float, Tuple, str] = 0.0,
+    latent_std: Union[float, Tuple, str] = 0.13025,
```
jazcollins, line 209: Should this be 1/0.13025?

jazcollins (commented):
looks good to me! left a comment about a default value

diffusion/models/models.py
coryMosaicML: Yes, it should. Nice catch! Fixed it now.

diffusion/models/models.py
```diff
@@ -31,13 +30,24 @@
 log = logging.getLogger(__name__)
 
 
+def _parse_latent_statistics(latent_stat: Union[float, Tuple, str]) -> Union[float, Tuple, str]:
+    if isinstance(latent_stat, str):
+        latent_stat = latent_stat.lower()
+        if latent_stat != 'latent_statistics':
+            raise ValueError(f'Invalid latent statistic {latent_stat}. Must be a float, tuple or "latent_statistics".')
+    elif type(latent_stat).__name__ == 'ListConfig' and not isinstance(latent_stat, float):
```
Skylion007, line 38: I feel like we can do OmegaConf.to_object() here?

diffusion/models/models.py
```diff
@@ -310,6 +332,11 @@ def stable_diffusion_xl(
                 layer = zero_module(layer)
         # Last conv block out projection
         unet.conv_out = zero_module(unet.conv_out)
+    if type(latent_mean) is float:
```
Skylion007, line 335: Technically should use isinstance of here

jazcollins (approved):

diffusion/models/models.py
coryMosaicML: Fixed!

PR #133, opened 2024-04-02
Arbitrary aspect ratio bucket boundaries

Landanjs:
Currently, a sample is assigned to the aspect ratio bucket with the smallest distance to the sample's aspect ratio bucket i.e. `torch.abs(self.aspect_ratio_buckets - orig_aspect_ratio).argmin()`. This results in implicit boundary conditions which are unlikely to optimal for reducing  overall max cropping and having similar max crops across buckets.

This PR adds an argument to specify the boundaries between aspect ratio buckets, allowing control over the max cropping across buckets.


coryMosaicML (approved):
Looks good to me! Thanks for the fix!

PR #134, opened 2024-04-02
Test PR

Landanjs:
Looks like PIL updated from `ANTIALIAS` to `LANCZOS` causing pyright failures in blank PRs. Here's the update, assuming we are okay with the package being updated? Wait, this is from a while ago, why is this happening now???


coryMosaicML (approved):
Looks good, thanks for fixing this.

PR #135, opened 2024-04-05
Only download CLIP on rank 0 when doing eval

coryMosaicML:
This PR has a fix to avoid multiple downloads of the CLIP model used to compute CLIP-FID in cleanfid. Also included is an optional additional pass through of additional args to `model.generate` in the evaluator.

### DIFF
--- a/diffusion/evaluation/clean_fid_eval.py
+++ b/diffusion/evaluation/clean_fid_eval.py
@@ -50,6 +50,7 @@ class CleanFIDEvaluator:
         precision (str): The precision to use for evaluation. Default: ``'amp_fp16'``.
         prompts (List[str], optional): The prompts to use for image visualtization.
             Default: ``["A shiba inu wearing a blue sweater]``.
+        additional_generate_kwargs (Dict, optional): Additional keyword arguments to pass to the model.generate method.
 
     """
 
--- a/diffusion/evaluation/clean_fid_eval.py
+++ b/diffusion/evaluation/clean_fid_eval.py
@@ -5,7 +5,7 @@
 
 import json
 import os
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 import clip
 import torch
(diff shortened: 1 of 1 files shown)


diffusion/evaluation/clean_fid_eval.py
```diff
@@ -107,7 +110,13 @@ def __init__(self,
...
 
         # Predownload the CLIP model for computing clip-fid
-        _, _ = clip.load('ViT-B/32', device=self.device)
+        clip_url = 'https://openaipublic.azureedge.net/clip/models/40d365715913c9da98579312b702a82c18be219cc2a73407c4526f58eba950af/ViT-B-32.pt'
+        clip_name = os.path.basename(clip_url)
+        clip_path = os.path.expanduser('~/.cache/clip')
+        if dist.get_local_rank() == 0:
+            clip.clip._download(clip_url, clip_path)
+        with dist.local_rank_zero_download_and_wait(os.path.join(clip_path, clip_name)):
+            _, _ = clip.load('ViT-B/32', device=self.device)
```
Landanjs, line 119: Nit
```suggestion
            clip.load('ViT-B/32', device=self.device)
```

diffusion/evaluation/clean_fid_eval.py
```diff
@@ -69,7 +70,8 @@ def __init__(self,
                  output_dir: str = '/tmp/',
                  num_samples: Optional[int] = None,
                  precision: str = 'amp_fp16',
-                 prompts: Optional[List[str]] = None):
+                 prompts: Optional[List[str]] = None,
+                 additional_generate_kwargs: Optional[Dict] = None):
```
Landanjs, line 74: Nit: any reason you aren't making this `generate_kwargs` and removing arguments from `__init__`? Not a big deal, it can always be done later.

Landanjs (approved):
LGTM! Just 2 nits, feel free to ignore

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: Fixed!

diffusion/evaluation/clean_fid_eval.py
coryMosaicML: The generate args that are taken from `__init__` raw are just the image size. Everything else has to get processed by the evaluator before seeing generate, so I'm inclined to leave the name as `additional_generate_kwargs` to avoid having to override the user specified values there.

PR #136, opened 2024-04-11
Bug to enable fp16

RR4787:
Latents are created based on default torch dtype but should be using dtype that the model is set to. Needed for inference


coryMosaicML (approved):
Thanks for catching this!

PR #138, opened 2024-04-16
Optional quasirandom timesteps, zero terminal SNR, cosine schedule for SD models

coryMosaicML:
This PR enables a few optional tweaks to the training process for SD style models that differ from the standard training recipe:

`quasirandomness`: Select timesteps quasi-randomly to reduce loss variance
`zero_terminal_snr`: Rescale the noise schedule to include zero SNR
`beta_schedule`: Allows one to select from `scaled_linear` (default), `linear`, or `squaredcos_cap_v2` noise schedules.

By default these are the standard values for SD2 and SDXL.


diffusion/models/stable_diffusion.py
```diff
@@ -140,13 +149,25 @@ def __init__(self,
...
     def _apply(self, fn):
         super(StableDiffusion, self)._apply(fn)
         self.latent_mean = fn(self.latent_mean)
         self.latent_std = fn(self.latent_std)
         return self
 
+    def _generate_quasirandom_timesteps(self, latents: torch.Tensor):
+        # Generate a quasirandom sequence of timesteps equal to the global batch size
+        global_batch_size = latents.shape[0] * dist.get_world_size()
+        timesteps = (len(self.noise_scheduler) * self.sobol_engine.draw(global_batch_size)).squeeze()
```
Landanjs, line 164: If this was 
```suggestion
        timesteps = ((len(self.noise_scheduler) - 1) * self.sobol_engine.draw(global_batch_size)).squeeze()
```

Would there be a need for the clamp below? To me, the clamp shouldn't be required so having it may mask issues with the code. I think I'd rather an assert if anything

diffusion/models/stable_diffusion.py
```diff
@@ -140,13 +149,25 @@ def __init__(self,
...
+        timesteps = torch.floor(timesteps).long().clamp(0, len(self.noise_scheduler) - 1)
```
Landanjs, line 165: Is there a reason for `floor()` instead of `round()`? I'm thinking `floor()` would add a slight bias toward lower timesteps, maybe?

diffusion/models/stable_diffusion.py
```diff
@@ -193,10 +214,13 @@ def forward(self, batch):
...
-                                  len(self.noise_scheduler), (latents.shape[0],),
-                                  device=latents.device,
-                                  generator=self.rng_generator)
+        if self.quasirandomness:
+            timesteps = self._generate_quasirandom_timesteps(latents)
+        else:
+            timesteps = torch.randint(0,
+                                      len(self.noise_scheduler), (latents.shape[0],),
+                                      device=latents.device,
+                                      generator=self.rng_generator)
```
Landanjs, line 223: Nit: to me, it would be cleaner to have `_generate_timesteps()` and place this if-statement within that function.

Landanjs (commented):
Nice! I had two questions about the quasirandom sampling, one clarification on the schedulers, and a nit. 

I'm not familiar with Sobol sequences, so I don't have an intuition if this is the best way to do it, but I trust your decision! From the description of Sobol sequences it sounds like what we need

diffusion/models/stable_diffusion.py
coryMosaicML: See response to the below comment for why it's `len(self.noise_scheduler)` instead of `len(self.noise_scheduler) - 1`

The clamp is because the sobol engine has a small bug where the output is occasionally `1.0` when it is supposed to be [0, 1.0). I don't think it has ever been fixed, but there is a years old bug report [here](https://github.com/pytorch/pytorch/issues/32047). In my testing this happens once every few million timestep generations, when it should happen with probability 0.

You're right that this might mask other issues though. In particular [this one](https://github.com/pytorch/pytorch/issues/33388) which would appear if we generate more than`2^30` timesteps this way. The clamp would still be necessary but I should probably add a check for this and reset the engine if this occurs.

coryMosaicML: A specific example of what I was thinking with `len(noise_scheduler) = 3`:

`rng = len(self.noise_scheduler) * self.sobol_engine.draw(global_batch_size)` is in [0, 3)
With `floor()`, one selects timesteps with the following boundaries:
```
0 if rng is in [0, 1.0)
1 if rng is in [1.0, 2.0)
2 if rng is in [2.0, 3.0)
```
Which gives equal probability to all timesteps.

With round, there would be an issue if `rng > 2.5` because that would give too many timesteps, but one could instead use the above suggested change and have 
`rng = (len(self.noise_scheduler) - 1) * self.sobol_engine.draw(global_batch_size)` be in [0, 2)
which with `round()` gives boundaries
```
0 if rng is in [0, 0.5)
1 if rng is in [0.5, 1.5)
2 if rng is in [1.5, 2.0)
```
But this is biased away from the 0 timestep, and the final timestep.

diffusion/models/stable_diffusion.py
coryMosaicML: See response to the below comment for why it's `len(self.noise_scheduler)` instead of `len(self.noise_scheduler) - 1`

The clamp is because the sobol engine has a small bug where the output is occasionally `1.0` when it is supposed to be [0, 1.0). I don't think it has ever been fixed, but there is a years old bug report [here](https://github.com/pytorch/pytorch/issues/32047). In my testing this happens once every few million timestep generations, when it should happen with probability 0.

You're right that this might mask other issues though. In particular [this one](https://github.com/pytorch/pytorch/issues/33388) which would appear if we generate more than`2^30` timesteps this way. The clamp would still be necessary but I should probably add a check for this and reset the engine if this occurs.

diffusion/models/models.py
```diff
@@ -145,14 +156,29 @@ def stable_diffusion_2(
...
+                                    beta_start=0.00085,
+                                    beta_end=0.012,
+                                    beta_schedule=beta_schedule,
+                                    trained_betas=None,
+                                    variance_type='fixed_small',
+                                    clip_sample=False,
+                                    prediction_type=prediction_type,
+                                    thresholding=False,
+                                    dynamic_thresholding_ratio=0.995,
+                                    clip_sample_range=1.0,
```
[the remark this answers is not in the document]
coryMosaicML: Yes, these are default values. I could just leave them out.

diffusion/models/stable_diffusion.py
Landanjs: I see! I didn't think about it enough, thanks!

diffusion/models/stable_diffusion.py
Landanjs: Thanks for the explanation!

Landanjs (approved):
LGTM

PR #141, opened 2024-05-09
Fix autoencoder load

RR4787:
Current autoencoder loading only works the first time. for inference if a model ever goes down and restarts, the load step is called again, but if overwrite isn't set it fails on creating the symlink cause it already exists.


coryMosaicML (approved):
Works for me, thanks!

PR #144, opened 2024-05-21
Script to pre-compute CLIP and T5

Landanjs:
Script to 
1. Stream a dataset containing text
2. Compute CLIP and T5 latents for the text
3. Save new latents with the previous dataset


coryMosaicML (approved):
Thanks for this!

PR #145, opened 2024-05-22
Script to generate images in bulk and optionally upload to cloud storage

coryMosaicML:
This PR adds an image generator that can be used to generate images in bulk across many GPUs. Currently only supports writing to local storage and uploading to OCI.


coryMosaicML:
Closing because this is a subset of [158](https://github.com/mosaicml/diffusion/pull/158).

PR #147, opened 2024-05-31
Fix bug in stable diffusion when mask_pad_tokens is false

coryMosaicML:
Currently when `mask_pad_tokens=False` in the stable diffusion model, the mask is still passed to the text encoder at generation time. This introduces a train/test mismatch because the masked tokens in the text encoder output are different from the unmasked tokens, and the model is trained with the unmasked pad tokens. The result is a degradation in image quality. The fix is to only use the mask from the tokenizer output if `mask_pad_tokens=True`.


Landanjs (approved):
LGTM!

PR #149, opened 2024-06-06
Only use a text encoder mask in SD model forward if mask_pad_tokens is false

coryMosaicML:
As is, we have a train/test mismatch when `mask_pad_tokens = False` where the mask is applied to the text encoder at generation time as expected by SDXL, but not during training time. This PR changes the train behavior to match the expected generation time behavior.


Landanjs (approved):
LGTM!

PR #155, opened 2024-07-09
MMDiT implementation and text-to-image training with rectified flows

coryMosaicML:
This PR contains an implementation of the MMDiT model from the [SD3 paper](https://arxiv.org/pdf/2403.03206), along with a model class for using it to train text to image models. To support this, a generic model inference class is also included. 

Major additions:
`diffusion/inference/inference_model.py` has `Modelnference` class for inference with arbitrary models from `models.py`
`diffusion/models/models.py` Includes a `text_to_image_transformer` model for SD3 style MMDiT
`diffusion/models/t2i_transformer.py` has the `ComposerModel` class for the MMDiT text to image model
`diffusion/models/transformer.py` has the layers/blocks for the MMDiT model
`diffusion/train.py` includes a new function to configure the optimizer for the new text to image model


diffusion/models/transformer.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def __init__(self, num_features: int, expansion_factor: int = 4):
+        super().__init__()
+        self.num_features = num_features
+        self.expansion_factor = expansion_factor
+        # AdaLN MLP for post-attention. Initialized to zero so modulation acts as identity at initialization.
+        self.adaLN_mlp_linear = nn.Linear(self.num_features, 4 * self.num_features, bias=True)
+        nn.init.zeros_(self.adaLN_mlp_linear.weight)
+        nn.init.zeros_(self.adaLN_mlp_linear.bias)
+        self.adaLN_mlp = nn.Sequential(nn.SiLU(), self.adaLN_mlp_linear)
```
A-Jacobson, line 203: probably refactor into something like 

```
AdaptiveLayerNorm(num_features, expansion_factor)
```
same for the one in PreAttentionBlock

diffusion/models/t2i_transformer.py
```diff
@@ -0,0 +1,459 @@
...
+        attention_flops = 4 * self.model.num_layers * seq_len**2 * self.model.num_features * batch_size
+        return 3 * param_flops + 3 * attention_flops
+
+    def encode_image(self, image: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Encode an image tensor with the autoencoder and patchify the latents."""
+        with torch.cuda.amp.autocast(enabled=False):
+            latents = self.autoencoder.encode(image.half())['latent_dist'].sample().data
+        # Scale and patchify the latents
+        latents = (latents - self.latent_mean) / self.latent_std
+        latent_patches, latent_coords = patchify(latents, self.patch_size)
```
A-Jacobson, line 215: I think it makes sense to move the patching/unpatching logic into the `DiffusionTransformer` class. This way the Transformer logic is contained to the transformer and we wouldn't have to pass `coords` to the forward pass. Thoughts?

diffusion/models/t2i_transformer.py
```diff
@@ -0,0 +1,459 @@
...
+            return attention_masks
+        elif len(attention_masks.shape) == 3:
+            encoder_attention_masks = attention_masks[:, 0]
+            for i in range(1, attention_masks.shape[1]):
+                encoder_attention_masks |= attention_masks[:, i]
+            return encoder_attention_masks
+        else:
+            raise ValueError(f'attention_mask should have either 2 or 3 dimensions: {attention_masks.shape}')
+
+    def make_text_embeddings_coords(self, text_embeddings: torch.Tensor) -> torch.Tensor:
```
A-Jacobson, line 255: Could this also happen in the `DiffusionTransformer` class?

A-Jacobson (approved):
Left a few commends about structure. Basically, I'd like to see more of the transformer logic confined to the transformer with the `ComposerModel` getting as close to "create the 3 models and call them" as possible. Overall, it's awesome and I conditionally approve as it's non-breaking and also has successful test runs.

diffusion/models/t2i_transformer.py
coryMosaicML: I'd like to leave patching/unpatching out of `DiffusionTransformer` as it's specific to doing text to image training without precomputed latents, and in principle `DiffusionTransformer` is agnostic to the modality/preprocessing.

diffusion/models/t2i_transformer.py
coryMosaicML: Same as above.

diffusion/models/transformer.py
coryMosaicML: Added an `AdaptiveLayerNorm` and also a `ModulationLayer` to replace the duplicated AdaLN logic in the Pre/Post attention blocks.

PR #157, opened 2024-07-12
Add option to use predefined aspect ratio buckets in the cropping transform

coryMosaicML:
This PR adds an additional cropping transformation that makes use of a dataset key specifying the desired aspect ratio of samples within a stream. It can be used by specifying `crop_type="bucketed_aspect_ratio"` and seething the value of `aspect_ratio_bucket_key`.


jazcollins (approved):
The logic looks good to me! 

I noticed a little typo in `RandomCropAspectRatioTransorm` and `RandomCropBucketedAspectRatioTransorm` (missing F in transform) so up to you if you wanna fix - doesn't affect anything because at least its consistent

PR #158, opened 2024-07-15
HF image generation that integrates with Cory's earlier script

rishab-partha:
Generates images at scale for an HF model from an HF dataset and patches those results to remote storage. Tested with 15 different HF models and with OCI.

### DIFF
--- a/diffusion/evaluation/generate_images.py
+++ b/diffusion/evaluation/generate_images.py
@@ -62,7 +62,7 @@ class ImageGenerator:
         if not hf_model:
             self.model = model
         else:
-            self.model = AutoPipelineForText2Image.from_pretrained(model, use_auth_token=True).to('cuda')
+            self.model = AutoPipelineForText2Image.from_pretrained(model, torch_dtype=torch.float16, use_auth_token=True).to('cuda')
         self.dataset = dataset
         self.load_path = load_path
         self.local_checkpoint_path = local_checkpoint_path
--- a/diffusion/evaluation/generate_images.py
+++ b/diffusion/evaluation/generate_images.py
@@ -13,15 +13,16 @@ from composer.core import get_precision_context
 from composer.utils import dist
 from composer.utils.file_helpers import get_file
 from composer.utils.object_store import OCIObjectStore
+from datasets import DatasetDict
 from diffusers import AutoPipelineForText2Image
 from torch.utils.data import Dataset
 from torchvision.transforms.functional import to_pil_image
 from tqdm.auto import tqdm
-from diffusers import AutoPipelineForText2Image
 
 
 class ImageGenerator:
     """Image generator that generates images from a dataset and saves them.
+
     Args:
         model (torch.nn.Module): The model to evaluate.
         dataset (Dataset): The dataset to use the prompts from.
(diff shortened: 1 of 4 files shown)


diffusion/evaluation/generate_images.py
```diff
@@ -0,0 +1,177 @@
...
+                 guidance_scale: float = 7.0,
+                 height: int = 1024,
+                 width: int = 1024,
+                 caption_key: str = 'caption',
+                 seed: int = 17,
+                 output_bucket: Optional[str] = None,
+                 output_prefix: Optional[str] = None,
+                 extra_keys: Optional[list] = None,
+                 additional_generate_kwargs: Optional[Dict] = None,
+                 hf_model: Optional[bool] = False):
```
coryMosaicML, line 59: Could you add this to the docstring?

diffusion/evaluation/generate_images.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if isinstance(model, str) and hf_model == False:
+            raise ValueError('Can only use strings for model with hf models!')
+        self.hf_model = hf_model
+        if hf_model or isinstance(model, str):
+            print(f'LOCALRANK{dist.get_local_rank()}')
```
coryMosaicML, line 65: Maybe a leftover print?

diffusion/evaluation/generate_images.py
```diff
@@ -0,0 +1,177 @@
...
+            if dist.get_local_rank() == 0:
```
coryMosaicML, line 66: Is this to ensure checkpoint download only happens on local rank 0, and the subsequent `from_pretrained` call doesn't trigger another download?

diffusion/evaluation/generate_images.py
```diff
@@ -0,0 +1,177 @@
...
+    def generate(self):
+        """Core image generation function. Generates images at a given guidance scale.
+
+        Args:
+            guidance_scale (float): The guidance scale to use for image generation.
+        """
+        os.makedirs(os.path.join('/tmp', self.output_prefix), exist_ok=True)
+        # Partition the dataset across the ranks
+        if self.hf_model:
+            dataset_len = self.dataset.num_rows  # type: ignore
```
coryMosaicML, line 120: Is this for using with a hf dataset?

diffusion/generate.py
```diff
@@ -0,0 +1,86 @@
...
+    if not config.hf_model:
+        model: ComposerModel = hydra.utils.instantiate(config.model)
+    else:
+        model = config.model.name
+
+    tokenizer = model.tokenizer if hasattr(model, 'tokenizer') else None
+
+    # The dataset to use for evaluation
+
+    if config.hf_model:
```
coryMosaicML, line 40: Is it possible to use hf datasets with non-hf models and vice versa? If so, using a `hf_model` flag to determine the dataset type might be confusing/limiting.

coryMosaicML (approved):
Looks good! Left a few minor comments/questions.

diffusion/evaluation/generate_images.py
rishab-partha: deleted

diffusion/evaluation/generate_images.py
rishab-partha: yup, also added to the dataset loading logic

diffusion/evaluation/generate_images.py
rishab-partha: yes, addressed in below comment

diffusion/generate.py
rishab-partha: added a new flag `hf_dataset` to delineate the difference

diffusion/evaluation/generate_images.py
rishab-partha: added

PR #164, opened 2024-08-10
Simple LoRA Finetuning (WIP)

rishab-partha:
Added a bit of simple code for LoRA finetuning diffusion models -- need to test with FSDP but this is minimal and runs with toy matrices on one GPU at the moment.

Works at tests with FSDP now -- 10 eps of FT on dataset of 760 images takes ~10 min with performance shown below, 
Sample image: Screenshot 2024-08-21 at 11 44 43 AM
Smoothed loss: 
Screenshot 2024-08-21 at 11 46 16 AM
Memory trace:
Screenshot 2024-08-21 at 11 47 30 AM


coryMosaicML (approved):
Looks good! Thanks for doing this and testing it out!

PR #165, opened 2024-08-20
Add option to change start and end SNR in SD2/SDXL configs

coryMosaicML:
This PR adds an option to adjust the starting/ending beta values for the stable diffusion noise schedules. It's useful when dealing with other datasets/autoencoders for which the default values may perform poorly.


jazcollins (approved):
LGTM, thanks!

PR #166, opened 2024-08-20
Pass loggers to Trainer in eval

jazcollins:
Pass loggers to Trainer and save it down - this is needed to initialize `MLFlowLogger` correctly

Remove initialization of `WandBLogger` so we don't double init it


coryMosaicML (approved):
LGTM! Thanks for fixing this!

PR #170, opened 2024-08-29
Update latent precomputation script with batching

jazcollins:
Update T5/CLIP latent precomputation script to use batching and add some light synthetic description preprocessing logic


coryMosaicML (approved):
Thanks for this update!

PR #171, opened 2024-09-11
Add composer model class for running with precomputed CLIP and T5 text latents

coryMosaicML:
This PR adds a model class for running with precomputed text latents. It is largely similar to the SDXL model, with the exception of concatenating latents from the different text encoders along a sequence dimension rather than a feature dimension after projecting to a common embedding dimension. 

This builds off of [152](https://github.com/mosaicml/diffusion/pull/152) which can now be closed.

### DIFF
--- a/diffusion/callbacks/log_diffusion_images.py
+++ b/diffusion/callbacks/log_diffusion_images.py
@@ -121,8 +121,9 @@ class LogDiffusionImages(Callback):
                 clip_attention_mask = clip_attention_mask.cpu().to(torch.long)
 
                 latent_batch['T5_LATENTS'] = t5_latents
+                latent_batch['T5_ATTENTION_MASK'] = t5_attention_mask
                 latent_batch['CLIP_LATENTS'] = clip_latents
-                latent_batch['ATTENTION_MASK'] = torch.cat([t5_attention_mask, clip_attention_mask], dim=1)
+                latent_batch['CLIP_ATTENTION_MASK'] = clip_attention_mask
                 latent_batch['CLIP_POOLED'] = clip_pooled
                 self.batched_latents.append(latent_batch)
 
@@ -144,11 +145,10 @@ class LogDiffusionImages(Callback):
             if self.precomputed_latents:
                 for batch in self.batched_latents:
                     pooled_prompt = batch['CLIP_POOLED'].cuda()
-                    prompt_mask = batch['ATTENTION_MASK'].cuda()
-                    t5_embeds = model.t5_proj(batch['T5_LATENTS'].cuda())
-                    clip_embeds = model.clip_proj(batch['CLIP_LATENTS'].cuda())
-                    prompt_embeds = torch.cat([t5_embeds, clip_embeds], dim=1)
-
+                    prompt_embeds, prompt_mask = model.prepare_text_embeddings(batch['T5_LATENTS'].cuda(),
+                                                                               batch['CLIP_LATENTS'].cuda(),
+                                                                               batch['T5_ATTENTION_MASK'].cuda(),
+                                                                               batch['CLIP_ATTENTION_MASK'].cuda())
                     gen_images = model.generate(prompt_embeds=prompt_embeds,
                                                 pooled_prompt=pooled_prompt,
                                                 prompt_mask=prompt_mask,
--- a/diffusion/callbacks/log_diffusion_images.py
+++ b/diffusion/callbacks/log_diffusion_images.py
@@ -39,6 +39,11 @@ class LogDiffusionImages(Callback):
         use_table (bool): Whether to make a table of the images or not. Default: ``False``.
         t5_encoder (str, optional): path to the T5 encoder to as a second text encoder.
         clip_encoder (str, optional): path to the CLIP encoder as the first text encoder.
+        t5_latent_key: (str): key to use for the T5 latents in the batch. Default: ``'T5_LATENTS'``.
+        t5_mask_key: (str): key to use for the T5 attention mask in the batch. Default: ``'T5_ATTENTION_MASK'``.
+        clip_latent_key: (str): key to use for the CLIP latents in the batch. Default: ``'CLIP_LATENTS'``.
+        clip_mask_key: (str): key to use for the CLIP attention mask in the batch. Default: ``'CLIP_ATTENTION_MASK'``.
+        clip_pooled_key: (str): key to use for the CLIP pooled in the batch. Default: ``'CLIP_POOLED'``.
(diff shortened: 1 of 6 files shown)


diffusion/datasets/image_caption_latents.py
```diff
@@ -140,6 +143,15 @@ def __getitem__(self, index):
                 if 'CLIP_LATENTS' in latent_key:
                     clip_pooled = np.frombuffer(sample[f'{caption_key}_CLIP_POOLED_TEXT'], dtype=np.float32).copy()
                     out['CLIP_POOLED'] = torch.from_numpy(clip_pooled).to(self.latent_dtype).reshape(latent_shape[1])
+        if self.drop_nans:
+            if out['CLIP_POOLED'].isnan().any():
```
jazcollins, line 147: this will work for us, but this section will fail if the text latent keys are named differently - probably safer to loop thru `self.text_latent_keys` and `self.attention_mask_keys`, e.g. something like:

``` 
if self.drop_nans:
    for latent_key, attn_key in zip(self.text_latent_keys, self.attention_mask_keys):
        if out[latent_key].isnan().any():
            out[latent_key] = torch.zeros_like(out[latent_key])
            out[attn_key] = torch.zeros_like(out[attn_key])
        if 'CLIP_LATENTS' in latent_key:
            out['CLIP_POOLED'] = torch.zeros_like(out['CLIP_POOLED'])
```

diffusion/callbacks/log_diffusion_images.py
```diff
@@ -121,8 +121,9 @@ def __init__(self,
                 clip_attention_mask = clip_attention_mask.cpu().to(torch.long)
 
                 latent_batch['T5_LATENTS'] = t5_latents
+                latent_batch['T5_ATTENTION_MASK'] = t5_attention_mask
```
jazcollins, line 124: ideally these key names wouldn't be hardcoded and would be grabbed from the dataset class but i think this OK for now

diffusion/models/models.py
```diff
@@ -580,6 +581,283 @@ def stable_diffusion_xl(
...
+    # Optionally load the tokenizers and text encoders
+    t5_tokenizer, t5_encoder, clip_tokenizer, clip_encoder = None, None, None, None
+    if include_text_encoders:
+        t5_tokenizer = AutoTokenizer.from_pretrained('google/t5-v1_1-xxl', cache_dir=cache_dir, local_files_only=True)
+        clip_tokenizer = AutoTokenizer.from_pretrained('stabilityai/stable-diffusion-xl-base-1.0',
+                                                       subfolder='tokenizer',
+                                                       cache_dir=cache_dir,
+                                                       local_files_only=False)
+        t5_encoder = AutoModel.from_pretrained('google/t5-v1_1-xxl',
+                                               torch_dtype=torch.float16,
```
jazcollins, line 786: hmm should be bfloat16? or read the dtype from an argument?

diffusion/models/models.py
```diff
@@ -580,6 +581,283 @@ def stable_diffusion_xl(
...
+                                               cache_dir=cache_dir,
+                                               local_files_only=False).encoder.eval()
+        clip_encoder = CLIPTextModel.from_pretrained('stabilityai/stable-diffusion-xl-base-1.0',
+                                                     subfolder='text_encoder',
+                                                     torch_dtype=torch.float16,
```
jazcollins, line 791: same comment about bfloat

diffusion/models/precomputed_text_latent_diffusion.py
```diff
@@ -0,0 +1,496 @@
...
+
+    def forward(self, batch):
+        latents, text_embeds, text_pooled_embeds, encoder_attention_mask = None, None, None, None
+
+        # Encode the images with the autoencoder encoder
+        inputs = batch['image']
+        latents = self.encode_images(inputs)
+
+        # Text embeddings are shape (B, seq_len, emb_dim), optionally truncate to a max length
+        t5_embed = batch['T5_LATENTS']
```
jazcollins, line 257: just to write comments consistent with what i said for the dataloader, probably safest to grab these key names from an argument but this is totally fine for our use cases

jazcollins (commented):
left some comments!

diffusion/datasets/image_caption_latents.py
coryMosaicML: This should be fixed now!

diffusion/models/models.py
coryMosaicML: Changed it to be an argument, but left the default as-is for inference.

diffusion/models/models.py
coryMosaicML: Same as above

diffusion/models/precomputed_text_latent_diffusion.py
coryMosaicML: Changed it to get key names from an argument.

diffusion/callbacks/log_diffusion_images.py
coryMosaicML: Changed these to come from arguments to the model class, but didn't yet make the change to grab them from the dataset class.

diffusion/models/models.py
```diff
@@ -580,6 +581,306 @@ def stable_diffusion_xl(
...
+            'stabilityai/stable-diffusion-xl-base-1.0'.
+        vae_model_name (str): Name of the VAE model to load. Defaults to
+            'madebyollin/sdxl-vae-fp16-fix' as the official VAE checkpoint (from
+            'stabilityai/stable-diffusion-xl-base-1.0') is not compatible with fp16.
+        autoencoder_path (optional, str): Path to autoencoder weights if using custom autoencoder. If not specified,
+            will use the vae from `model_name`. Default `None`.
+        autoencoder_local_path (optional, str): Path to autoencoder weights. Default: `/tmp/autoencoder_weights.pt`.
+        include_text_encoders (bool): Whether to include text encoders in the model. Should only do this for running
+            inference. Default: `False`.
+        text_encoder_dtype (str): The dtype to use for the text encoder. One of [`float32`, `float16`, `bfloat16`].
```
jazcollins, line 628: Perhaps the default value for this should match the default value of `latent_dtype` of `StreamingImageCaptionLatentsDataset`? (diffusion/datasets/image_caption_latents.py on L44 which sets it to bfloat)

jazcollins (approved):
Left one comment, otherwise LGTM!

diffusion/models/models.py
coryMosaicML: Fixed!

